PR #17, opened 2020-07-01
stellar-etl: Set up file for account export command

### LINKED ISSUE #8: cmd: set up accounts command
We want to set up an accounts command, which will eventually get accounts in a specified timerange. This will live in `cmd/accounts.go` and correspond to the `accounts` table in BigQuery.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

A file for a command line function that exports accounts. The file contains comments describing the eventual functionality and flags for the command. Closes #8 

### Why

This is one of the commands that the ETL will use for exporting data. 

### Known limitations

The command is not functional, just described.

### DIFF
--- a/cmd/accounts.go
+++ b/cmd/accounts.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// accountsCmd represents the accounts command
+var accountsCmd = &cobra.Command{
+	Use:   "accounts",
+	Short: "Exports the account data.",
+	Long:  `Exports historical account data within the specified ledger range to an output file.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		/*
+			Functionality planning:
+			1. Read in start and end ledger numbers
+			2. Provide ledger number to ingestion system and receive data
+			3. Convert data from ingestion into an array of Ledger structs with length equal to the limit
+			4. Write array to output file
+			TODO: Consider having a way to export directly to BigQuery using bigquery library
+		*/
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(accountsCmd)
+	/*
+		Needed flags:
+			TODO: determine if providing ledger sequence number or timestamp is preferable (possibly could do both)
+
+			start-time: the time for the beginning of the period to export; default to genesis ledger's creation time
+			end-time: the time for the end of the period to export (required)
+
+			start-ledger: the ledger sequence number for the beginning of the export period
+			end-ledger: the ledger sequence number for the end of the export range (required)
+
+			limit: maximum number of accounts to export; default to 1000
+			output-file: filename of the output file
+	*/
+}


cmd/accounts.go
```diff
@@ -0,0 +1,39 @@
...
+	Short: "Exports the account data.",
+	Long:  `Exports historical account data within the specified ledger range to an output file.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		/*
+			Functionality planning:
+			1. Read in start and end ledger numbers
+			2. Provide ledger number to ingestion system and receive data
+			3. Convert data from ingestion into an array of Ledger structs with length equal to the limit
+			4. Write array to output file
+			TODO: Consider having a way to export directly to BigQuery using bigquery library
```
debnil, line 19: Good foresight thinking about the rest of the project. Note that we want to use the Google Cloud Composer (their managed service build above Airflow) for this part. We would likely use an Airflow task for this, since there are custom Airflow operators for BigQuery ([example](https://airflow.apache.org/docs/stable/_api/airflow/contrib/operators/bigquery_operator/index.html)). So, exporting to BigQuery doesn't have to be part of this CLI tool.

cmd/accounts.go
```diff
@@ -0,0 +1,39 @@
...
+		*/
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(accountsCmd)
+	/*
+		Needed flags:
+			TODO: determine if providing ledger sequence number or timestamp is preferable (possibly could do both)
```
debnil, line 28: To figure this out, we can take a look at the ingestion engine code, likely the `HistoryArchiveBackend`, and figure out what metadata can be included in the structs that it outputs.

cmd/accounts.go
```diff
@@ -0,0 +1,39 @@
...
+
+			start-time: the time for the beginning of the period to export; default to genesis ledger's creation time
```
debnil, line 30: This is an interesting default. I'm not sure I agree with it, since a user could easily accidentally kick it off from the genesis ledger, rather than just the last five minutes. I don't think we need to change this comment yet, because that might be the right use case. This will become clearer in the next few weeks as we implement more.

cmd/accounts.go
```diff
@@ -0,0 +1,39 @@
...
+			end-time: the time for the end of the period to export (required)
+
+			start-ledger: the ledger sequence number for the beginning of the export period
+			end-ledger: the ledger sequence number for the end of the export range (required)
+
+			limit: maximum number of accounts to export; default to 1000
```
debnil, line 36: We should empirically measure this maximum value. I think we need to make sure that we can fully cover a five minute period with one call to the CLI tool. 

Some back of the envelope math for the worst case. Since ledgers close every five seconds, there should be 60 ledgers in that period. A ledger contains up to 50 transactions, and a transaction can contain up to 100 operations. Thus, there are up to 60 x 50 x 100 = 300,000 operations in five minutes. It is possible, though incredibly improbable, that all of these operations alter account state (i.e., [`Create Account`](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#create-account), [`Change Trust`](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#change-trust), [`Allow Trust`](https://www.stellar.org/developers/guides/concepts/list-of-operations.html#allow-trust)).

PR #47, opened 2020-07-14
internal: Added transform function for ledgers

### LINKED ISSUE #35: Implement transform function for ledger data
### What
We need to implement a function that transforms the information we get about ledgers from the new ingestion system into a form suitable for BigQuery. Part of #33.

### Why
This function will be used when exporting ledger information from the CLI.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR adds the transform function for ledgers. The function takes in a ledger in the format described by the ingestion system, and transforms it into a custom struct suitable for BigQuery. Closes #35 

### Why

The transform function will be part of the larger CLI tool. This function will allow us to create a command that retrieves ledgers and exports them in the form we want.


internal/transform/ledger.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/stellar/go/xdr"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+//ConvertLedger converts a ledger from the history archive ingestion system into a form suitable for BigQuery
+func ConvertLedger(inputLedgerMeta xdr.LedgerCloseMeta) (LedgerOutput, error) {
+	ledger, ok := inputLedgerMeta.GetV0()
+	if !ok {
+		return LedgerOutput{}, errors.New("Could not access the version 0 information for the provided ledger")
```
debnil, line 16: Preference for `fmt.Errorf` instead of `errors.New`, though it matters less when you have no formatting. I'd also make the error message a bit shorter, since the developer likely has context: "could not access v0 info for given ledger". It would be even better if you could print out the ledger's sequence number in the error, so the developer could more easily debug.

internal/transform/ledger.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	ledgerHeaderHistory := ledger.LedgerHeader
+	ledgerHeader := ledgerHeaderHistory.Header
+
+	outputSequence := int32(ledgerHeader.LedgerSeq)
+	if outputSequence < 0 {
+		return LedgerOutput{}, errors.New("The sequence is a negative value")
```
debnil, line 23: The sequence is somewhat vague. Maybe "ledger sequence %d is negative"? It would be helpful if there was some other metadata in that error to identify the problematic ledger as well, maybe a timestamp?

internal/transform/ledger.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	outputCloseTime, err := utils.TimePointToUTCTimeStamp(ledgerHeader.ScpValue.CloseTime)
+	if err != nil {
+		return LedgerOutput{}, err
+	}
+
+	outputTotalCoins := int64(ledgerHeader.TotalCoins)
+	if outputTotalCoins < 0 {
+		return LedgerOutput{}, errors.New("The total number of coins is a negative value")
```
debnil, line 47: Add the ledger seqnum to this error and below messages for easier debugging.

internal/transform/ledger.go
```diff
@@ -0,0 +1,127 @@
...
+		if results[i].Result.Successful() {
+			successTxCount++
+			operationCount += numberOfOps
+		} else {
+			failedTxCount++
+		}
+
+	}
+
+	return int32(txCount) - failedTxCount, operationCount, successTxCount, failedTxCount, strconv.FormatInt(int64(txSetOperationCount), 10), nil
```
debnil, line 126: Function looks good, but define the variables you're returning above the `return` statement. It'll be much more readable and help with debugging too.

internal/transform/ledger_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func createTestArchiveBackend() *ledgerbackend.HistoryArchiveBackend {
+	archiveStellarURL := "http://history.stellar.org/prd/core-live/core_live_001"
+	backend, err := ledgerbackend.NewHistoryArchiveBackendFromURL(archiveStellarURL)
+	utils.PanicOnError(err)
+	return backend
+}
+func TestTransformOnTrueLedger(t *testing.T) {
+	sequenceNumber := uint32(30578981)
+	backend := createTestArchiveBackend()
```
debnil, line 26: I should've made this clearer, but in unit testing, don't establish a network connection or get live data. This is generally because the file that lives at a specific URL can change, and introducing a live network connection adds a source of significant network latency to your testing. You'd like to be able to run the tests as quickly as possible.

I get the impetus in this case to add the extra test, but since you're also testing these functions with hand-crafted test cases below, the only additional information is whether or not the `ledgerbackend` struct is working as expected. But this information would be handled in a separate phase of the pipeline.

internal/transform/ledger_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestTransformLedger(t *testing.T) {
+	type transformTest struct {
+		input         xdr.LedgerCloseMeta
+		desiredOutput LedgerOutput
+		desiredError  error
+	}
+
+	tests := []transformTest{
```
debnil, line 70: Great use of table-driven testing! :)

internal/transform/ledger_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	convertedLedger, err := ConvertLedger(testLedger)
+	assert.Equal(t, desiredOutput, convertedLedger)
+
+}
+
+func TestTransformLedger(t *testing.T) {
+	type transformTest struct {
+		input         xdr.LedgerCloseMeta
+		desiredOutput LedgerOutput
```
debnil, line 66: Change desiredOutput to wantOutput, as per Go convention.

internal/transform/ledger_test.go
```diff
@@ -0,0 +1,153 @@
...
+		desiredError  error
```
debnil, line 67: Similarly, change desiredError to wantErr, as per Go convention.

internal/utils/main.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"encoding/hex"
+	"errors"
+	"time"
+
+	"github.com/stellar/go/xdr"
+)
+
+//PanicOnError is a function that panics if the provided error is not nil
```
debnil, line 11: I like the instincts, but I'm not actually sure this is the best practice in Go. @leighmcculloch I feel like you'd know the answer to this.

internal/transform/schema.go
```diff
@@ -0,0 +1,38 @@
+package transform
+
+import "time"
+
+//LedgerOutput is a representation of a ledger that aligns with the BigQuery table history_ledger
+type LedgerOutput struct {
+	//sequence number of the ledger
```
debnil, line 7: Put the comment to the right of the field you're referring to, as it reduces ambiguity.

internal/transform/schema.go
```diff
@@ -0,0 +1,38 @@
+package transform
+
+import "time"
+
+//LedgerOutput is a representation of a ledger that aligns with the BigQuery table history_ledger
```
debnil, line 5: Struct looks great :)

internal/transform/ledger_test.go
```diff
@@ -0,0 +1,153 @@
...
+		LedgerHeader: xdr.LedgerHeaderHistoryEntry{
+			Header: header,
+		}})
+	utils.PanicOnError(err)
+	return lcm
+}
+
+func createSampleTx(sequence uint32) xdr.TransactionEnvelope {
+	kp, err := keypair.Random()
+	if err != nil {
```
debnil, line 127: If you're going to create a `PanicOnError` function, may as well use it everywhere :)

internal/utils/main.go
```diff
@@ -0,0 +1,32 @@
...
+	return hexString
+}
+
+//TimePointToUTCTimeStamp takes in an xdr TimePoint and converts it to a time.Time struct in UTC. It returns an error for negative timepoints
+func TimePointToUTCTimeStamp(providedTime xdr.TimePoint) (time.Time, error) {
+	intTime := int64(providedTime)
+	if intTime < 0 {
+		return time.Now(), errors.New("The timepoint is negative")
+	}
+	return time.Unix(intTime, 0).UTC(), nil
```
debnil, line 31: Why does it need the extra `UTC()` function call?

Isaiah-Turner: Without it, time.Unix defaults to EDT (might just be my timezone). We stored everything in the database in UTC

PR #50, opened 2020-07-18
internal: added transform for operations

### LINKED ISSUE #37: Implement transform function for operations data
### What
We need to implement a function that transforms the information we get about operations from the new ingestion system into a form suitable for BigQuery. Part of #33.

### Why
This function will be used when exporting operations data from the CLI.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR adds the transform function for operations. The function takes in an operation in the format described by the ingestion system, and transforms it into a custom struct suitable for BigQuery. Closes #37 
### Why

The transform function will be part of the larger CLI tool. This function will allow us to create a command that retrieves operations and exports them in the form we want.

### DIFF
--- a/internal/transform/operation.go
+++ b/internal/transform/operation.go
@@ -1,0 +1,350 @@
+package transform
+
+import (
+	"encoding/base64"
+	"fmt"
+	"strconv"
+
+	"github.com/stellar/go/amount"
+	ingestio "github.com/stellar/go/exp/ingest/io"
+	"github.com/stellar/go/xdr"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+//TransformOperation converts an operation from the history archive ingestion system into a form suitable for BigQuery
+func TransformOperation(operation xdr.Operation, operationIndex int32, transaction ingestio.LedgerTransaction) (OperationOutput, error) {
+	outputSourceAccount, err := utils.GetAccountAddressFromMuxedAccount(getOperationSourceAccount(operation, transaction))
+	if err != nil {
+		return OperationOutput{}, err
+	}
+
+	outputOperationType := int32(operation.Body.Type)
+	if outputOperationType < 0 {
+		return OperationOutput{}, fmt.Errorf("The operation type (%d) is negative for  operation %d", outputOperationType, operationIndex)
+	}
+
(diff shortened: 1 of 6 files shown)


internal/transform/operation.go
```diff
@@ -0,0 +1,345 @@
...
+	operationType := operation.Body.Type
+	allOperationResults, ok := transaction.Result.OperationResults()
+	if !ok {
+		return Details{}, fmt.Errorf("Could not access any results for this transaction")
+	}
+
+	currentOperationResult := allOperationResults[operationIndex]
+	switch operationType {
+	case xdr.OperationTypeCreateAccount:
+		op := operation.Body.MustCreateAccountOp()
```
debnil, line 152: I'm wary of the `Must` here and below. Seems like it might panic rather than returning an error message, which would help explain where the user should re-ingest.

internal/transform/operation.go
```diff
@@ -0,0 +1,345 @@
...
+		shouldAuth := xdr.TrustLineFlags(op.Authorize).IsAuthorized()
+		outputDetails.Authorize = shouldAuth
+
+	case xdr.OperationTypeAccountMerge:
+		aid := operation.Body.MustDestination().ToAccountId()
+		outputDetails.Account = sourceAccountAddress
+		outputDetails.Into = aid.Address()
+
+	case xdr.OperationTypeInflation:
+		// inflation is deprecated
```
debnil, line 327: This needs to be handled, since you'll be ingesting operations from genesis, and you don't want to just drop all the inflation ops.

internal/transform/operation.go
```diff
@@ -0,0 +1,345 @@
...
+	return transformedOperation, nil
+}
+
+func getOperationSourceAccount(operation xdr.Operation, transaction ingestio.LedgerTransaction) xdr.MuxedAccount {
+	sourceAccount := operation.SourceAccount
+	if sourceAccount != nil {
+		return *sourceAccount
+	}
+
+	return transaction.Envelope.SourceAccount()
```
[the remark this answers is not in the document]
Isaiah-Turner: I need in the account id in the allow trust case when I'm gettin the operation details. If I have the MuxedAccount, it's easy to get the id.

PR #51, opened 2020-07-20
transform: added transform for accounts

### LINKED ISSUE #34: Implement transform function for account data
### What
We need to implement a function that transforms the information we get about accounts from the new ingestion system into a form suitable for BigQuery. Part of #33.

### Why
This function will be used when exporting account information from the CLI.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR adds the transform function for accounts. The function takes in an account in the format described by the ingestion system, and transforms it into a custom struct suitable for BigQuery. Closes #34.
### Why

The transform function will be part of the larger CLI tool. This function will allow us to create a command that retrieves accounts and exports them in the form we want.


internal/transform/account.go
```diff
@@ -0,0 +1,93 @@
+package transform
+
+import (
+	"fmt"
+
+	"github.com/stellar/go/xdr"
+)
+
+//TransformAccount converts an account from the history archive ingestion system into a form suitable for BigQuery
```
debnil, line 9: Nit: Put a space between `//` and the function name.

internal/transform/account.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return AccountOutput{}, err
+	}
+
+	outputBalance := int64(accountEntry.Balance)
+	if outputBalance < 0 {
+		return AccountOutput{}, fmt.Errorf("Balance is negative (%d) for account: %s", outputBalance, outputID)
+	}
+
+	accountExtentionInfo, V1Found := accountEntry.Ext.GetV1()
```
debnil, line 26: Worth adding a comment here explaining what the `V1` struct contains. In the case of a bug in this section, that wouldn't be immediately obvious, and the information would help debugging.

debnil, line 26: Nit: change Extention to Extension.

PR #64, opened 2020-07-27
transform: fixed type conversions

### LINKED ISSUE #52: Ensure that type casts don't lose information
### What
Some type casts in the transform package may be invalid. Casting from uint32 to int32 could result in a loss of information, since uint32s have a higher maximum value. Similarly, casting from int64 to int32 could result in a loss of information. We need to ensure these casts do not happen, likely by changing the type of the field in the output struct to ensure it can hold all the information.
### Why
Losing data in type casts makes the ETL transform functions unreliable. We want to have an accurate and complete picture of the data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR makes fixes to type conversions that could have resulted in a loss of data. Now all type conversions should be from xdr wrapper types to their corresponding built-in go types. Closes #52.

### Why
We don't want the transform functions to lose data. The ETL output should be correct and consistent with the ingestion system data.


internal/transform/account.go
```diff
@@ -45,9 +45,6 @@ func TransformAccount(ledgerEntry xdr.LedgerEntry) (AccountOutput, error) {
 	}
 
 	outputNumSubentries := uint32(accountEntry.NumSubEntries)
-	if outputNumSubentries < 0 {
```
debnil, line 48: Good catch with the uints :)

internal/transform/ledger.go
```diff
@@ -53,27 +50,16 @@ func TransformLedger(inputLedgerMeta xdr.LedgerCloseMeta) (LedgerOutput, error)
...
-		return LedgerOutput{}, fmt.Errorf("The base reserve (%d) is negative for ledger %d", outputBaseReserve, outputSequence)
-	}
+	outputBaseReserve := uint32(ledgerHeader.BaseReserve)
 
-	outputMaxTxSetSize := int32(ledgerHeader.MaxTxSetSize)
-	if outputMaxTxSetSize < 0 {
-		return LedgerOutput{}, fmt.Errorf("The max transaction set size (%d) is negative for ledger %d", outputMaxTxSetSize, outputSequence)
-	} else if outputMaxTxSetSize < outputTransactionCount {
+	outputMaxTxSetSize := uint32(ledgerHeader.MaxTxSetSize)
+	if int64(outputMaxTxSetSize) < int64(outputTransactionCount) {
```
debnil, line 58: Why the cast here for value comparison?

Isaiah-Turner: outputMaxTxSetSize is a uint32 and outputTransactionCount is an int32. By casting to an int64, both of their respective data ranges are captured and they can be compared safely.

PR #67, opened 2020-07-28
stellar-etl: implement export_ledgers command

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements the export_ledgers command. It allows users to connect to the history archives, retrieve ledger information in a desired range, and export the JSON encodings to an output file.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table history_ledgers more efficiently.

### Known limitations


cmd/export_ledgers_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+func TestTransformLedger(t *testing.T) {
+	type cliTest struct {
+		name    string
+		args    []string
+		golden  string
```
debnil, line 42: Great use of golden files :)

cmd/export_ledgers.go
```diff
@@ -0,0 +1,122 @@
...
+			marshalled, err := json.Marshal(transformed)
+			if err != nil {
+				logger.Fatal(fmt.Sprintf("could not json encode ledger %d: ", startNum+uint32(i)), err)
+			}
+
+			outFile.Write(marshalled)
+			outFile.WriteString("\n")
+		}
+
+		/*
```
debnil, line 87: I think you can delete these planning comments -- looks all implemented :)

cmd/export_ledgers.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(ledgersCmd)
+	ledgersCmd.Flags().Uint32P("start-ledger", "s", 0, "The ledger sequence number for the beginning of the export period")
+	ledgersCmd.Flags().Uint32P("end-ledger", "e", 0, "The ledger sequence number for the end of the export range (required)")
+	ledgersCmd.Flags().Uint32P("limit", "l", 60, "Maximum number of ledgers to export")
+	ledgersCmd.Flags().StringP("output-file", "o", "exported_ledgers.txt", "Filename of the output file")
+	ledgersCmd.MarkFlagRequired("end-ledger")
+	/*
```
debnil, line 105: The commented flags should be modified to account for what you've developed :)

PR #68, opened 2020-07-29
stellar-etl: implement export_transactions command

### LINKED ISSUE #59: cmd: implement command for exporting transactions
### What
We need to implement the planned functionality for exporting transactions. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated transaction data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements the export_transactions command. It allows users to connect to the history archives, retrieve transaction information in a desired range, and export the JSON encodings to an output file. Closes #59.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table history_transactions more efficiently.

### Known limitations


cmd/export_ledgers.go
```diff
@@ -62,7 +67,7 @@ var ledgersCmd = &cobra.Command{
 		// TODO: check the permissions of the file to ensure that it can be written to
 		outFile, err := os.OpenFile(absolutePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
 		if err != nil {
-			logger.Fatal("could not create output file: ", err)
+			logger.Fatal("could not open output file: ", err)
```
debnil, line 70: This error message should communicate which command is having the problem. But there's a chance that it isn't an open issue, so this message could be deceptive. Suggestion: error in output file for <command>:

cmd/export_ledgers_test.go
```diff
@@ -85,38 +79,48 @@ func TestTransformLedger(t *testing.T) {
...
+		// By reading the error text from the logger, we can extract the lower level error that the user would see
+		if test.golden == "" {
+			errorMsg := fmt.Errorf(extractErrorMsg(string(errText)))
+			assert.Equal(t, test.wantErr, errorMsg)
+		}
+
+		// Tests that are designed to fail won't generate an valid output file, and so testing shouldn't look for one
+		if test.golden != "" {
+			assert.Equal(t, test.wantErr, actualError)
+			actualOutput, err := ioutil.ReadFile(testFileName)
```
debnil, line 114: Thinking more, there are two viable routes to testing the output.
1. What you're doing here: write to a file, read that file back in testing, check it against a reference.
2. Add a CLI flag to write the output to stdout, check that output in memory.

I think 2 is better for testing, since it avoids potential memory issues in a future CI system. For this PR, this is fine. I would just add a TODO to add a flag for an output option; implement stdout; and update tests accordingly.

cmd/export_transactions.go
```diff
@@ -0,0 +1,84 @@
...
+
+// transactionsCmd represents the transactions command
+var transactionsCmd = &cobra.Command{
+	Use:   "export_transactions",
+	Short: "Exports the transaction data over a specified range.",
+	Long:  `Exports the transaction data over a specified range to an output file.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		startNum, endNum, limit, path := getBasicFlags(cmd)
+
+		absolutePath, err := filepath.Abs(path)
```
debnil, line 22: This code is well-structured, so adding an option to check if path != nil should be easy, and quickly enable the stdout behavior discussed above :)

PR #69, opened 2020-07-30
stellar-etl: implement export_operations command

### LINKED ISSUE #57: cmd: implement command for exporting operations
### What
We need to implement the planned functionality for exporting operations. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated operation data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements the export_operations command. It allows users to connect to the history archives, retrieve operations information in a desired range, and export the JSON encodings to an output file. Closes #57.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table history_operations more efficiently.

### Known limitations


internal/input/operations.go
```diff
@@ -0,0 +1,65 @@
...
+					OperationIndex: int32(index),
+					Transaction:    tx,
+				})
+
+				if uint32(len(opSlice)) >= limit {
+					break
+				}
+			}
+		}
+		if uint32(len(opSlice)) >= limit {
```
debnil, line 59: I think there's a very possible bug that would arise from this usage. If the developer passes a start and end sequence number, they'd assume that the limit would be the difference between those numbers. But since the default value is 60, this would result in a bunch of ledgers being dropped!

I think the solution is simple: within `getBasicFlags`, set limit as max(end - start, limit); but curious what your thoughts are.

Also, nit: newline after close brace.

PR #73, opened 2020-08-06
stellar-etl: implement export_accounts command

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements some of the export_accounts command. It allows users to connect to the history archives, retrieve account information starting from the genesis ledger, and exports the JSON encodings to an output file. 

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table history_accounts more efficiently.

### Known limitations
Right now, the bucket list is used to get all the information starting from the genesis ledger, but it's difficult to get arbitrary ranges. In the next PR, the command will use captive core in order to update information in arbitrary ranges. 


cmd/export_accounts.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/stellar/stellar-etl/internal/transform"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+var accountsCmd = &cobra.Command{
+	Use:   "export_accounts",
+	Short: "Exports the account data.",
+	Long:  `Exports historical account data within the specified range to an output file.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		startNum, endNum, limit, path, useStdOut := utils.MustBasicFlags(cmd.Flags(), cmdLogger)
```
debnil, line 19: Naming convention: `stdout` is a phrase, so the variable should be named `useStdout`.

cmd/export_accounts.go
```diff
@@ -0,0 +1,68 @@
...
+				fmt.Println(string(marshalled))
+			}
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(accountsCmd)
+	utils.AddBasicFlags("accounts", accountsCmd.Flags())
+	ledgersCmd.MarkFlagRequired("end-ledger")
```
debnil, line 55: Not sure I follow this. Why are you modifying `ledgersCmd`? Or should this be `accountsCmd`?

internal/input/accounts.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/stellar/go/exp/ingest/adapters"
+	ingestio "github.com/stellar/go/exp/ingest/io"
+	"github.com/stellar/go/historyarchive"
+	"github.com/stellar/go/xdr"
+)
+
+// GetAccounts returns a slice of ledger close metas for the ledgers in the provided range (inclusive on both ends)
+func GetAccounts(start, end uint32, limit int64) ([]xdr.LedgerEntry, error) {
+	archiveStellarURL := "http://history.stellar.org/prd/core-live/core_live_001"
```
debnil, line 16: Pull this out of the function as a package-wide variable - it'll be helpful for other commands.

internal/input/accounts.go
```diff
@@ -0,0 +1,79 @@
...
+
+		return readBucketList(archive, checkpointSeq, xdr.LedgerEntryTypeAccount)
+	}
+
+	// Use captive core to read a range otherwise
+	// TODO: connect to captive core and read from there
+	return readFromCaptive(start, end)
+}
+
+func readFromCaptive(start, end uint32) ([]xdr.LedgerEntry, error) {
```
debnil, line 51: Nit: Add a TODO in the function body, indicating it needs to be implemented.

internal/input/accounts.go
```diff
@@ -0,0 +1,79 @@
...
+	return []xdr.LedgerEntry{}, nil
+}
+
+func readBucketList(archive *historyarchive.Archive, checkpointSeq uint32, entryType xdr.LedgerEntryType) ([]xdr.LedgerEntry, error) {
```
debnil, line 55: Function looks great! Awesome job thinking through effective use of the bucket list. Very readable and understandable :)

cmd/export_accounts.go
Isaiah-Turner: Good point. That should be accountsCmd.

PR #75, opened 2020-08-10
stellar-etl: finish export_accounts command

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR finalizes the export_accounts command. Unneeded flags were removed and the help messages were improved.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table history_accounts more efficiently.

### Known limitations
The end-time flag is not implemented. The ingestion system uses sequence numbers instead of times. If it is expanded to use times, then we can add the flags. The serialize-method flag also is not implemented, since at the moment we only have json encodings for the structs.


internal/input/accounts.go
```diff
@@ -0,0 +1,69 @@
...
+
+	checkpointSeq, err := utils.GetCheckpointNum(end, latestNum)
+	if err != nil {
+		return []xdr.LedgerEntry{}, err
+	}
+
+	return readBucketList(archive, checkpointSeq, xdr.LedgerEntryTypeAccount)
+}
+
+func readBucketList(archive *historyarchive.Archive, checkpointSeq uint32, entryType xdr.LedgerEntryType) ([]xdr.LedgerEntry, error) {
```
debnil, line 45: I'd consider pulling out this function into a separate file, as other CLI's might also need it. That can be done in a subsequent PR for sure.

PR #76, opened 2020-08-10
cmd: Set up file for ledger entry changes export command

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

A file for a command line function that exports accounts, trustlines, and offers from the history archive. The comments in the file detail the desired functionality and flags.

### Why

This is one of the commands handles stellar core and the subcommands that rely on it. Planning out its functionality is important because this is a core part of the CLI.

### Known limitations

The command is not functional, just described.

### DIFF
--- a/cmd/export_ledger_entry_changes.go
+++ b/cmd/export_ledger_entry_changes.go
@@ -17,6 +17,7 @@ package cmd
 
 import (
 	"github.com/spf13/cobra"
+	"github.com/stellar/stellar-etl/internal/utils"
 )
 
 // exportLedgerEntryChangesCmd represents the exportLedgerEntryChanges command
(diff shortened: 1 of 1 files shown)


cmd/export_ledger_entry_changes.go
```diff
@@ -0,0 +1,78 @@
+/*
```
debnil, line 1: Remove the license from the file, not needed.

cmd/export_ledger_entry_changes.go
```diff
@@ -0,0 +1,78 @@
...
+limitations under the License.
+*/
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+// exportLedgerEntryChangesCmd represents the exportLedgerEntryChanges command
```
debnil, line 23: This comment is a little repetitive, try to expand it to be more descriptive of actual function.

cmd/export_ledger_entry_changes.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(exportLedgerEntryChangesCmd)
+	utils.AddBasicFlags("changes", exportLedgerEntryChangesCmd.Flags())
+	exportLedgerEntryChangesCmd.Flags().Uint32P("batch-size", "b", 64, "number of ledgers to export changes from in each batches")
+	exportLedgerEntryChangesCmd.Flags().BoolP("export-accounts", "a", false, "set in order to export account changes")
+	exportLedgerEntryChangesCmd.Flags().BoolP("export-trustlines", "t", false, "set in order to export trustline changes")
+	exportLedgerEntryChangesCmd.Flags().BoolP("export-offers", "f", false, "set in order to export offer changes")
+
```
debnil, line 58: How are you planning to handle trades? I realize they're not a direct entry change, so it might be tricky.

cmd/export_ledger_entry_changes.go
Isaiah-Turner: Oh, that comment was autogenerated when I made the command. I'll remove it

cmd/export_ledger_entry_changes.go
Isaiah-Turner: I'll need to look at what the TxMeta provides. If it gives trade information that we need, then we probably will have to parse each ledger to get changes and transactions.

PR #77, opened 2020-08-11
stellar-etl: goroutine and channels for ledger changes export

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What

This PR implements reading data from stellar core and sending it to the correct channel. It also supports reading data from the channels.

### Why

In order to export data that is represented as ledger entries, we need a long running stellar core instance that runs in the background. By setting up these channels, we can access information from the core instance as it appears.

### Known limitations

Entries that are deleted are not represented. In addition, no information is exported yet, its just stored in memory. As a result, some flags are not used.

### DIFF
--- a/cmd/export_ledger_entry_changes.go
+++ b/cmd/export_ledger_entry_changes.go
@@ -5,10 +5,8 @@ import (
 	"path/filepath"
 
 	"github.com/spf13/cobra"
-	ingestio "github.com/stellar/go/exp/ingest/io"
 	"github.com/stellar/go/xdr"
 	"github.com/stellar/stellar-etl/internal/input"
-	"github.com/stellar/stellar-etl/internal/transform"
 	"github.com/stellar/stellar-etl/internal/utils"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/export_ledger_entry_changes.go
```diff
@@ -18,6 +25,90 @@ confirmed by the Stellar network.
...
+					}
+				}
+			}
+
+			if accChannel == nil && offChannel == nil && trustChannel == nil {
+				break
+			}
+		}
+
+		fmt.Println(transformedAccounts)
```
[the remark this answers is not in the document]
debnil: To be specific: I think this will work in a bounded setting, but when unbounded, we likely want to apply some batching strategy. Otherwise, I don't think we'd ever get printed output, but we would eventually get an out-of-memory error once these arrays of changes get too large.

internal/input/changes.go
```diff
@@ -0,0 +1,148 @@
...
+	if trustChannel != nil {
+		close(trustChannel)
+	}
+}
+
+// StreamChanges reads in ledgers, processes the changes, and send the changes to the channel matching their type
+func StreamChanges(core *ledgerbackend.CaptiveStellarCore, start, end uint32, accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
+	if end != 0 {
+		for seq := start; seq <= end; seq++ {
+			changeReader, err := ingestio.NewLedgerChangeReader(core, network.PublicNetworkPassphrase, seq)
```
debnil, line 116: Pull out this inner for loop into a separate function, and call it in each of the branches of this `if-else`.

cmd/export_ledger_entry_changes.go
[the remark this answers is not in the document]
Isaiah-Turner: Yeah, this was just for testing bounded ranges. I should have put a TODO here for the next PR.

cmd/export_ledger_entry_changes.go
```diff
@@ -43,18 +48,13 @@ be exported.`,
...
-		}*/
-
 		core, err := input.PrepareCaptiveCore(execPath, configPath, startNum, endNum)
 		if err != nil {
 			cmdLogger.Fatal("error creating a prepared captive core instance: ", err)
 		}
 
 		accChannel, offChannel, trustChannel := createChangeChannels(exportAccounts, exportOffers, exportTrustlines)
-		go input.StreamChanges(core, startNum, endNum, accChannel, offChannel, trustChannel)
+		input.StreamChanges(core, startNum, endNum, accChannel, offChannel, trustChannel)
```
debnil, line 57: This function needs to be run in the background as a goroutine, otherwise it'll be blocking, and you'll never get to the below code. The `go` is necessary to kick that off.

debnil: Ah, I see, you made the function run a goroutine. I'm not sure that'll prevent blocking, but I can believe it works. In either case, I think it's better to kick off the goroutine in the CLI - it'll be clearer and more readable.

cmd/export_ledger_entry_changes.go
```diff
@@ -18,6 +25,97 @@ confirmed by the Stellar network.
...
+		}
+
+		accChannel, offChannel, trustChannel := createChangeChannels(exportAccounts, exportOffers, exportTrustlines)
+		input.StreamChanges(core, startNum, endNum, accChannel, offChannel, trustChannel)
+
+		transformedAccounts := make([]transform.AccountOutput, 0)
+		transformedOffers := make([]transform.OfferOutput, 0)
+		transformedTrustlines := make([]transform.TrustlineOutput, 0)
+
+		for {
```
debnil, line 63: I completely miswrote this earlier. Refactor this code so the blocking `for-select` is in a separate function (`input.ReceiveChanges`), that function should return the three slices and an error. This will be functionally the same but significantly clean up the command-line code.

PR #81, opened 2020-08-17
stellar-etl: implement export_ledger_entry_changes command

### LINKED ISSUE #54: cmd: implement command for exporting accounts
### What
We need to implement the planned functionality for exporting accounts. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated account data.

### LINKED ISSUE #56: cmd: implement command for exporting offers
### What
We need to implement the planned functionality for exporting offers. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated offer data.

### LINKED ISSUE #60: cmd: implement command for exporting trustlines
### What
We need to implement the planned functionality for exporting trustlines. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated trustline data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements the export_ledger_entry_changes command. This command reads ledger changes, which includes the accounts, offers, and trustlines. These changes are exported in batches within a user-specified range. The range can be bounded or unbounded. Closes #54. Closes #56. Closes #60.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery tables accounts, offers, and trust_lines.

### Known limitations
Deleted entries are not handled. In a subsequent PR, we need to update the transform functions to handle deleted entries so that we know which rows to remove from the BigQuery tables.

The end-time flag is not implemented. The ingestion system uses sequence numbers instead of times. If it is expanded to use times, then we can add the flags. The serialize-method flag also is not implemented, since at the moment we only have json encodings for the structs.


internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
-		for _, change := range compactedChanges {
-			// TODO: figure out how to handle deleted entries
-			if change.Post != nil {
-				entry := *change.Post
-				sendToChannel(entry, accChannel, offChannel, trustChannel)
+func addLedgerChangesToCache(changeReader *ingestio.LedgerChangeReader, accCache, offCache, trustCache *ingestio.LedgerEntryChangeCache) error {
+	for {
+		change, err := changeReader.Read()
+		if err == ingestio.EOF {
+			break
```
debnil, line 117: Why not just `return nil` here?

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
+		case xdr.LedgerEntryTypeAccount:
+			accCache.AddChange(change)
+
+		case xdr.LedgerEntryTypeOffer:
+			offCache.AddChange(change)
+
+		case xdr.LedgerEntryTypeTrustline:
+			trustCache.AddChange(change)
+
+		default:
```
debnil, line 134: Is there a `default` case not covered by the above cases? If so and it's a rare event, might be worth adding some logging here. If so and it's not rare, add a comment explaining this case.

cmd/export_ledger_entry_changes.go
```diff
@@ -52,49 +66,78 @@ be exported.`,
...
+		go input.StreamChanges(core, startNum, endNum, batchSize, accChannel, offChannel, trustChannel, cmdLogger)
+		if endNum != 0 {
+			batchCount := int(math.Ceil(float64(endNum-startNum+1) / float64(batchSize)))
+			for i := 0; i < batchCount; i++ {
+				transformedAccounts, transformedOffers, transformedTrustlines := input.ReceiveChanges(accChannel, offChannel, trustChannel, cmdLogger)
+				exportTransformedData(outFile, useStdout, transformedAccounts, transformedOffers, transformedTrustlines)
+			}
+
+		} else {
+			batchNum := uint32(1)
```
debnil, line 79: Nit: slightly more idiomatic to do `var batchNum uint32 = 1`

cmd/export_ledger_entry_changes.go
```diff
@@ -52,49 +66,78 @@ be exported.`,
...
+			for {
```
debnil, line 80: I think this is an infinite loop? When would this terminate?

internal/input/changes.go
```diff
@@ -12,6 +13,16 @@ import (
 	"github.com/stellar/stellar-etl/internal/utils"
 )
 
+const password = network.PublicNetworkPassphrase
+
+// ChangeBatch represents the changes in a batch of ledgers represented by the range [BatchStart, BatchEnd)
```
debnil, line 18: Nice work representing a batch -- good and minimal struct :)

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
 			}
+
+			err = addLedgerChangesToCache(changeReader, accChanges, offChanges, trustChanges)
+			if err != nil {
+				logger.Error(fmt.Sprintf("unable to read changes from ledger %d: ", seq), err)
+			}
+
+			changeReader.Close()
+
+			// only look at the next ledger if this one was available; otherwise try again until the network closes the ledger
```
debnil, line 165: I get the intention of this comment, but I don't understand why this code would retry. My understanding: it would make a new reader (if the last observed seqnum is before the latest ledger); cache the changes as needed; close the reader; and then increment the seqnum.

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
+
+	}
+
+	accBatch := ChangeBatch{
+		Changes:    accChanges.GetChanges(),
+		BatchStart: batchStart,
+		BatchEnd:   batchEnd,
+		Type:       xdr.LedgerEntryTypeAccount,
 	}
+	sendBatchToChannels(accBatch, accChannel, offChannel, trustChannel)
```
debnil, line 177: Since you know `ChangeBatch.Type`, I would make the channels you're not sending the batch to `nil`, to avoid any accidental serialization bugs.

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
 
-			acc, err := transform.TransformAccount(entry)
-			if err != nil {
-				errorMsg := fmt.Sprintf("error transforming account entry last updated at: %d", entry.LastModifiedLedgerSeq)
-				logger.Error(errorMsg, err)
-				break
-			}
+			for _, change := range batch.Changes {
+				// TODO: export deleted changes as well as created/updated entries
+				if change.Post == nil {
```
debnil, line 238: It's unclear what `change.Post` means. An explanatory comment would help future editors avoid diving through `xdr.LedgerEntryChange` documentation to understand this check.

A generally useful principle (and, to be clear, a personal opinion): if a check requires understanding internals of an imported package, it's good to err towards more explanatory comments, rather than solely relying on a clever and clean-looking implementation.

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
 	transformedTrustlines := make([]transform.TrustlineOutput, 0)
-
+	accBatchRead, offBatchRead, trustBatchRead := false, false, false
 	for {
-
 		select {
-		case entry, ok := <-accChannel:
+		case batch, ok := <-accChannel:
 			if !ok {
 				accChannel = nil
```
debnil, line 232: It might be useful to explain why we set the channel as `nil` because reading one entry failed.

internal/input/changes.go
```diff
@@ -116,97 +110,205 @@ func closeChannels(accChannel, offChannel, trustChannel chan xdr.LedgerEntry) {
...
+				entry := *change.Post
+				acc, err := transform.TransformAccount(entry)
+				if err != nil {
+					errorMsg := fmt.Sprintf("error transforming account entry last updated at: %d", entry.LastModifiedLedgerSeq)
+					logger.Error(errorMsg, err)
+					break
+				}
+
+				transformedAccounts = append(transformedAccounts, acc)
+				accBatchRead = true
```
debnil, line 251: I think you're setting this flag as true after reading a single `Change` from each of the panels, rather than from reading an entire batch. It's possible I'm wrong, but if not, make sure to change this, since it's breaking.

cmd/export_ledger_entry_changes.go
Isaiah-Turner: This is the unbounded case where exports should continue happening indefinitely as the network updates.

PR #83, opened 2020-08-18
stellar-etl: implement export_trades command

### LINKED ISSUE #58: [1] cmd: implement command for exporting trades
### What
We need to implement the planned functionality for exporting trades. The command will use the `transform` package functions for transforming data. Part of #53.

### Why
This is one step towards having a functional pipeline for data transfer. The command will provide users with a way to access regularly updated trade data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR implements the export_trades command. It allows users to connect to the history archives, retrieve trades information in a desired range, and export the JSON encodings to an output file. Closes #58.

### Why
This command is part of the larger CLI tool. It will allow for updating the BigQuery table trades more efficiently.

### Known limitations


cmd/export_trades.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(tradesCmd)
+	utils.AddBasicFlags("trades", tradesCmd.Flags())
+	tradesCmd.MarkFlagRequired("end-ledger")
+
+	/*
```
debnil, line 62: Either remove this comment; edit it to only include TODOs; or move it above the `tradesCmd` declaration, so it's easier for the user to understand the command's semantics.

internal/input/trades.go
```diff
@@ -0,0 +1,97 @@
...
+			tx, err := txReader.Read()
+			if err == ingestio.EOF {
+				break
+			}
+
+			for index, op := range tx.Envelope.Operations() {
+				// Trades occur on these operation types: manage buy offer, manage sell offer,
+				// create passive sell offer, path payment send, and path payment receive
+				// Trades also can only occur when these operations are successful
+				if operationResultsInTrade(op) && tx.Result.Successful() {
```
debnil, line 59: These are necessary conditions, but are these sufficient for an operation to be a trade? For example, wouldn't this mean a successful offer operation would count as a trade?

cmd/export_trades.go
```diff
@@ -0,0 +1,76 @@
...
+			cmdLogger.Fatal("could not read trades: ", err)
+		}
+
+		for _, tradeInput := range trades {
+			trades, err := transform.TransformTrade(tradeInput.OperationIndex, tradeInput.Transaction, tradeInput.CloseTime)
+			if err != nil {
+				cmdLogger.Fatal("could not transform trade ", err)
+			}
+
+			// Each operation, when transformed, results in multiple trades, each of which needs to be exported
```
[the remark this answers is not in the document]
Isaiah-Turner: From how operation results are set up, each operation can have multiple trades. The parent transaction holds the results for each operation, but trades are only visible within the operation results.

internal/input/trades.go
Isaiah-Turner: Not all of these operations will result in trades, but this is checked in TransformTrade (an empty slice is returned if no trades occurred) for a given op.

PR #87, opened 2020-08-19
stellar-etl: add offer ids to trades

### LINKED ISSUE #84: [0.5] Add base and counter offer ids to trades
### What
We need to get the base and counter offer ids for trades. If the counter offer id is not available, we need to generate one in a manner identical to the way Horizon generates ids.

### Why
These ids are part of the existing BigQuery tables, so we want to keep them. If ids are generated, they need to work the same way as the horizon ids to ensure ecosystem wide consistency.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds the base and counter offer id fields to the export_trades command. Closes #84.

### Why
The trades tables in BigQuery has both of these fields, but they weren't added in the first pass of the transform trade function. This PR adds them in to remain consistent with the data that is already in BigQuery.


debnil:
I'll review this once #83 is merged to `master`, and this branch is rebased on the new `master`. Right now, any review of this would be redundant.

PR #91, opened 2020-08-24
stellar-etl: implement time range to ledger range cmd

### LINKED ISSUE #89: [1] Add get_ledger_range_from_times command
### What
We need to add a command that converts a time range into a ledger range.

### Why
We get execution times for our Airflow runs, but the CLI takes in ledger ranges. This command will convert times to ranges, which will be passed to our export tasks.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds implements the command for converting a time range into a ledger range. Closes #89.

### Why
When using Airflow to manage the exports, we can only retrieve the execution time for a given task. We need to convert this time into a ledger range that can be used by the export commands.

### Known limitations


internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type graphPoint struct {
+	Seq       int64
+	CloseTime time.Time
+}
+
+type graph struct {
+	Backend              *ledgerbackend.HistoryArchiveBackend
+	BeginPoint, EndPoint graphPoint
```
debnil, line 18: Nit: Place struct fields, even if the same type, on different lines for cleaner readability.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	if currentLedger > 1 {
+		tempLedger := currentLedger - 1
+		tempTime, err := g.getGraphPoint(tempLedger)
+		if err != nil {
+			return 0, err
+		}
+
+		if (tempTime.CloseTime.Before(targetTime) || tempTime.CloseTime.Equal(targetTime)) && (currentTime.CloseTime.After(targetTime) || currentTime.CloseTime.Equal(targetTime)) {
```
debnil, line 106: It might be easier to cast the three times you're comparing to their epoch time (using `.Unix()`). You can then directly compare integral values via standard arithmetic signs.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	defer graph.close()
+
+	err = graph.limitLedgerRange(&startTime, &endTime)
+	if err != nil {
+		return 0, 0, err
+	}
+
+	startLedger, err := graph.findLedgerForDate(2, startTime)
```
debnil, line 44: It'll be useful to add a comment explaining the hardcoded 2.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return graph, err
+	}
+
+	graph.EndPoint = latestPoint
+	return graph, nil
+}
+
+// findLedgerForDate recursively searches for the ledger that was closed on or directly after targetTime
+func (g graph) findLedgerForDate(currentLedger int64, targetTime time.Time) (int64, error) {
```
debnil, line 93: Note: it's fine to use a value receiver given the concise nature of the `graph` struct. But [here](https://tour.golang.org/methods/8) is a useful link to the Golang tour motivating that choice.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/stellar/go/exp/ingest/ledgerbackend"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+type graphPoint struct {
```
debnil, line 11: It'll be helpful to add a comment explaining this struct (though un-exported), given how heavily it's used below. This will help guide future maintenance and explain initial design decisions.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+	Seq       int64
+	CloseTime time.Time
+}
+
+type graph struct {
```
debnil, line 16: As with `graphPoint`, it'll be helpful to add a comment explaining this struct as well.

internal/input/ledger_range.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// findLedgerForDate recursively searches for the ledger that was closed on or directly after targetTime
+func (g graph) findLedgerForDate(currentLedger int64, targetTime time.Time) (int64, error) {
+	currentTime, err := g.getGraphPoint(currentLedger)
+	if err != nil {
+		return 0, err
+	}
+
+	if currentLedger > 1 {
```
debnil, line 99: It seems a little strange to have the base case nested in an `if`-statement, and after an additional calculation. I would decrement the currentLedger before passing it to the recursive call, and then the base case here can be simpler.

PR #99, opened 2020-09-02
transform: converted nil slices to empty slices

### LINKED ISSUE #98: [2] Change null slice to empty slices
### What
There are places in the transform where slices are used. We need to ensure that when a slice is empty, the outputted field is an empty slice instead of `null`.

### Why
BigQuery treats an empty array differently from `null`. If we have columns with null, it will cause errors when loading data into BigQuery.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR changes nil slices to empty slices. When exporting as json, these slices are now `[]` instead of `null`. Closes #98.

### Why
Fields with `null` values cause problems when loading into BigQuery. They are treated as strings instead of empty arrays. This needs to be corrected at the CLI level.

### Known limitations


internal/transform/operation.go
```diff
@@ -421,5 +421,28 @@ func extractOperationDetails(operation xdr.Operation, transaction ingestio.Ledge
...
+
+	ensureSlicesAreNotNil(&outputDetails)
 	return outputDetails, nil
 }
+
+func ensureSlicesAreNotNil(details *Details) {
+	if details.Path == nil {
+		details.Path = make([]AssetOutput, 0)
+	}
+	if details.SetFlags == nil {
```
debnil, line 433: Nit: newline.

debnil (approved):
LGTM!

PR #105, opened 2020-09-18
stellar-etl: change transform errors severity

### LINKED ISSUE #104: [5] Change transformation errors to warnings
### What
When the stellar-etl throws an error in the transformation process, it closes fatally. It isn't possible to continue going once the error has been thrown. Instead, the default behavior should be to throw a warning that indicates what a transform failed and continue processing the rest of the entries. An optional flag `strict-export` will allow users to change the ETL so that it exits fatally once an error is hit.

### Why
There are edge cases where the transform fails even though the input is correct. For example, when a transaction fails, we still attempt to transform its operations. If the transform on this operation fails, exiting the program does not help the user. There is no way for them to continue, and retrying the run will not help. By changing the default behavior, we will allow users to get the rest of the correct data even if there is some incorrect data.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR reduces the default severity of transform failures to warnings. It also introduces a flag that will return transform failures to their old severity (fatal). Closes #104.

### Why
The old default behavior was very brittle. We would like a higher failure tolerance for the stellar-etl by default, with the option to treat errors more seriously with CLI flags
 
### Known limitations

### DIFF
--- a/cmd/export_trades.go
+++ b/cmd/export_trades.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/stellar/stellar-etl/internal/toid"
+
 	"github.com/spf13/cobra"
 	"github.com/stellar/stellar-etl/internal/input"
 	"github.com/stellar/stellar-etl/internal/transform"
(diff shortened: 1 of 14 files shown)


cmd/export_trades.go
```diff
@@ -29,17 +32,34 @@ var tradesCmd = &cobra.Command{
 			cmdLogger.Fatal("could not read trades: ", err)
 		}
 
+		failures := 0
 		for _, tradeInput := range trades {
 			trades, err := transform.TransformTrade(tradeInput.OperationIndex, tradeInput.OperationHistoryID, tradeInput.Transaction, tradeInput.CloseTime)
 			if err != nil {
-				cmdLogger.Fatal("could not transform trade ", err)
+				parsedID := toid.Parse(tradeInput.OperationHistoryID)
+				locationString := fmt.Sprintf("from ledger %d, transaction %d, operation %d", parsedID.LedgerSequence, parsedID.TransactionOrder, parsedID.OperationOrder)
```
debnil, line 40: Very clean representation. Just FYI - you could potentially use "%v+" - but I think this is cleaner.

debnil (approved):
LGTM! I'm a big fan of the pattern you're using here to display failures. It's a really clean division between failure and warning, and the CLI API is ergonomic. It's really showing your improvement and growth as an engineer -- great stuff!

PR #107, opened 2020-10-01
stellar-etl: fix infinite loops

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds a binary search as a backup when converting from times to ledger ranges. If an infinite cycle occurs in the original algorithm, the binary search is used.

### Why
The original algorithm is faster than a binary search because it uses properties of the network to find the correct ledger. However, it isn't guaranteed to find the ledger, unlike binary search. Using binary search as a backup ensures that the command will always terminate.
 
### Known limitations


debnil (approved):
LGTM! Nice fix :)

PR #110, opened 2020-10-27
stellar-etl: removed exp from ingestion import paths

Isaiah-Turner:
### What
This PR removes `exp/` from the import path for ingestion modules.
 
### Why
Ingestion modules were promoted from experimental in [this PR](https://github.com/stellar/go/pull/3118). The old import paths are no longer valid, which was stopping the build process.
 
### Known limitations 
N/A


debnil (approved):
LGTM!

PR #111, opened 2020-10-29
stellar-etl: add export_orderbooks cmd

### LINKED ISSUE #108: [5](previous-3) Add export_orderbooks command
### What
Add a command to export order books. The command should be able to export raw order books or normalized files that fit the BigQuery schemas defined in stellar/voyager#18.

### Why
Adding the new command will support the work on liquidity ledgers. See stellar/voyager#18 for information about liquidity ledgers.

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds the export_orderbooks command to the ETL. The command exports normalized orderbooks. Closes #108.

### Why
This command allows us to export historical orderbooks, which is vital in calculating liquidity in the network's past.

### Known limitations 
N/A


cmd/export_ledger_entry_changes_test.go
```diff
@@ -5,9 +5,11 @@ import (
 	"testing"
 )
 
+const coreExecutablePath = "/Users/isaiahturner/Local/stellar-core/src/stellar-core"
```
debnil, line 8: Replace local directories with arbitrary filepaths that'll work for anyone who clones the repo.

cmd/export_orderbooks.go
```diff
@@ -0,0 +1,151 @@
...
+		orderbook, err := input.GetEntriesFromGenesis(checkpointSeq, xdr.LedgerEntryTypeOffer)
+		if err != nil {
+			cmdLogger.Fatal("could not read initial orderbook: ", err)
+		}
+
+		input.UpdateOrderbook(checkpointSeq, startNum, orderbook, core, cmdLogger)
+		orderbookChannel := make(chan input.OrderbookBatch)
+
+		go input.StreamOrderbooks(core, startNum, endNum, batchSize, orderbookChannel, orderbook, cmdLogger)
+		if endNum != 0 {
```
debnil, line 70: It'd be useful to add a comment explaining the significance of `endNum == 0`, since it's unclear from just looking at the code.

Other than that, very clean and clear implementation of this function!

cmd/export_orderbooks.go
```diff
@@ -0,0 +1,151 @@
...
+				if batchEnd > endNum {
+					batchEnd = endNum
+				}
+
+				parser := input.ReceiveParsedOrderbooks(orderbookChannel, strictExport, cmdLogger)
+				exportOrderbook(batchStart, batchEnd, folderPath, useStdout, strictExport, parser)
+			}
+		} else {
+			var batchNum uint32 = 0
+			for {
```
debnil, line 85: What prevents this infinite loop from blocking? Is there any termination condition?

A comment briefly covering the above would be handy.

cmd/export_orderbooks.go
```diff
@@ -0,0 +1,151 @@
...
+				batchEnd := batchStart + batchSize - 1
+				parser := input.ReceiveParsedOrderbooks(orderbookChannel, strictExport, cmdLogger)
+				exportOrderbook(batchStart, batchEnd, folderPath, useStdout, strictExport, parser)
+				batchNum++
+			}
+		}
+	},
+}
+
+func writeSlice(file *os.File, useStdout bool, slice [][]byte) {
```
debnil, line 96: This function is slightly difficult to parse because of the double `if` statements, but it's generally quite clean. So, I think that's better addressed through an explanatory comment at the top, than a less readable implementation.

internal/input/orderbooks.go
```diff
@@ -0,0 +1,258 @@
...
+)
+
+type OrderbookBatch struct {
+	BatchStart uint32
+	BatchEnd   uint32
+	Orderbooks map[uint32][]ingestio.Change
+}
+
+// OrderbookParser handles parsing orderbooks
+type OrderbookParser struct {
```
debnil, line 22: If there's any official Stellar documentation relevant to understanding this data structure (i.e., structure of orderbook), it would be great to reference that here in a comment.

internal/input/orderbooks.go
```diff
@@ -0,0 +1,258 @@
...
+			marshalledMarket, err := json.Marshal(converted.Market)
+			if err != nil {
+				errorMsg := fmt.Sprintf("error json marshalling market for offer: %d", converted.Offer.HorizonID)
+				if o.Strict {
+					o.Logger.Fatal(errorMsg, err)
+				} else {
+					o.Logger.Warning(errorMsg, err)
+					continue
+				}
+			} else {
```
debnil, line 85: I might be wrong, but I think you don't need this in an `else` statement, since `Fatal` or `continue` would both move to the next iteration and not append the `marshalledMarket`.

This applies to a few other `else` clauses below.

internal/input/orderbooks.go
```diff
@@ -0,0 +1,258 @@
...
+		seq++
+	}
+
+	return offChanges, nil
+}
+
+func exportOrderbookBatch(batchStart, batchEnd uint32, core *ledgerbackend.CaptiveStellarCore, orderbookChan chan OrderbookBatch, startOrderbook []ingestio.Change, logger *log.Entry) {
+	batchMap := make(map[uint32][]ingestio.Change)
+	batchMap[batchStart] = make([]ingestio.Change, len(startOrderbook))
+	copy(batchMap[batchStart], startOrderbook)
```
debnil, line 161: Why the copy over assigment? My understanding is `make(map...)` allows for dynamic sizing of values, so you shouldn't need to make then copy. Could be totally off, since I can't find direct documentation.

i.e., `batchMap[batchStart] = startOrderbook`

internal/input/orderbooks.go
```diff
@@ -0,0 +1,258 @@
...
+		batchStart := start
+		batchEnd := batchStart + batchSize
+		for {
+			exportOrderbookBatch(batchStart, batchEnd, core, orderbookChannel, startOrderbook, logger)
+			batchStart = batchEnd
+			batchEnd = batchStart + batchSize
+		}
+	}
+
+	//close(orderbookChannel)
```
debnil, line 225: Delete commented channel

debnil (approved):
LGTM! Some organizational comments, but generally great work :)

internal/input/orderbooks.go
Isaiah-Turner: There isn't any documentation right now. I'll open another PR to update the readme with the exact schema/definition.

internal/input/orderbooks.go
Isaiah-Turner: The orderbook gets modified later on as new changes are applied to the orderbook. If we don't copy values, then the value in the map ends up changing as well, since its just a reference.

PR #118, opened 2020-12-16
cmd: change to truncate exported files

Isaiah-Turner:
### What
This PR truncates exported files instead of appending.

### Why
Sometimes the Airflow setup would write to the same file. With the append mode, this meant that data was duplicated in the file. We need to ensure that a file only contains one exported batch.


debnil (approved):
LGTM!

PR #121, opened 2020-12-18
stellar-etl: add export_assets cmd

Isaiah-Turner:
<details>
  
### PR Structure

  otherwise).
  packages.

### Thoroughness

</details>

### What
This PR adds the export_assets command to the ETL. The command exports assets from payment operations.

### Why
This command allows us to export assets with a content-based hash as an id.

### Known limitations 
N/A


cmd/export_assets.go
```diff
@@ -0,0 +1,105 @@
...
+				} else {
+					cmdLogger.Warning(errMsg, err)
+					failures++
+					continue
+				}
+
+			}
+
+			// if we have seen the asset already, do not export it
+			if _, exists := seenIDs[transformed.AssetID]; exists {
```
debnil, line 50: I'm a little wary of an in-memory map... Do we know that this will be accessible when ingesting from history archives and then when ingesting using captive-core? Might we invoke this command on different bounds of ledgers? We need total state of all assets for this check to work. Else, it definitely will have duplicates.

That said, I think this can be solved at the Airflow/BigQuery level through merges. Just flagging that at the ETL level, I'm not positive this works.

internal/input/assets.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	opSlice := []OperationTransformInput{}
+	for seq := start; seq <= end; seq++ {
+		txReader, err := ingestio.NewLedgerTransactionReader(backend, publicPassword, seq)
+		if err != nil {
+			return []OperationTransformInput{}, err
+		}
+
+		for int64(len(opSlice)) < limit || limit < 0 {
```
debnil, line 35: Why are we handling `limit < 0` or `limit >=0` in this loop? This seems like it's passed in and not modified in this loop. So, we can just early return if `limit < 0` before we even enter this loop. This doesn't change any function, but it does result in a cleaner approach with fewer conditions.

cmd/export_assets.go
debnil: In any case, might be good to add a comment here explaining why an in-memory map is an appropriate and sufficient internal data structure.

debnil (commented):
Looks good so far. Few comments, but nothing structural. Let me know when you want a final review :)

cmd/export_assets.go
Isaiah-Turner: You're right that overall we will have duplicate values. This in-memory map only ensures that there won't be duplicate assets within a given export. When we merge in Airflow, if an asset shows up multiple times in a single file, it will be inserted into the table multiple times. This means we need to ensure there is never duplicated info within a single export. From there, Airflow merges can handle the rest.

internal/input/assets.go
Isaiah-Turner: This loop is reading each transaction at the ledger. Limit isn't changed, but if the length of the slice reaches limit, then we want to stop reading. Similarly, within a transaction there can be multiple operations, so we need that second check while iterating over the operations in the tx.Envelope.

cmd/export_assets.go
debnil: Cool - this makes sense to me :)

internal/input/assets.go
debnil: Ah, I was a bit unclear. I agree that we need the second check, for the reasons you stated. I'm not sure why we need to check `limit < 0`.

debnil (approved):
LGTM! One outstanding question, but structure looks good.

internal/input/assets.go
Isaiah-Turner: If we have a negative limit, then that means we export everything. This statement is saying while `we are under the limit`, or `the limit is negative/we want to export everything`, keep exporting.

PR #123, opened 2021-06-18
update go version, update Dockerfile

acharb:
updating the golang version to 1.16 as part of this security fix: https://github.com/stellar/security/issues/42

also updating the Dockerfile steps

note: I also ran `go mod tidy`


docker/Dockerfile
```diff
@@ -1,13 +1,15 @@
-FROM golang:1.14
-ENV GOBIN /go/bin
-RUN go get github.com/stellar/stellar-etl
```
acharb, line 3: Running the original Dockerfile gave me build errors. I think these changes is what we want anyways though. It follows the pattern of other Go Dockerfiles we have

Makefile
```diff
@@ -5,7 +5,7 @@ SUDO := $(shell docker version >/dev/null 2>&1 || echo "sudo")
 ETLHASH=stellar/stellar-etl:$(shell git rev-parse --short HEAD)
 
 docker-build:
-	$(SUDO) docker build --no-cache -t $(ETLHASH) -t stellar/stellar-etl:latest ./docker
+	$(SUDO) docker build --no-cache -t $(ETLHASH) -t stellar/stellar-etl:latest -f ./docker/Dockerfile .
```
acharb, line 8: needed this change for the changes in Dockerfile

docker/Dockerfile
```diff
@@ -1,13 +1,15 @@
...
+FROM golang:1.16
+
+COPY . /etl
```
acharb, line 3: we weren't installing from local before. I think that's why it's giving build errors of trying fetch remotely. Not sure if that was intentional, but I think these changes are what we want anyways though. It follows the pattern of other Go Dockerfiles we have

nikhilsaraf (approved):
LGTM, thanks! 🎉

PR #124, opened 2021-08-03
Szwise update ingest package

sydneynotthecity:
`stellar-etl` pinned a version of the `stellar/go` library from Jun 2020. Because of this, it was missing new updates to the protocols as well as new Operations added in Protocol 11+. Updates include upgrading the `stellar/go` version to the latest (2021 Jul release), refactoring the code to adjust for `stellar/go` upgrade and parsing out the additional Operations.

*Changes*
* pin version `v0.0.0-20210730151625-babac4f8a0b7` of  the Stellar go monorepo
* introduce `context` to all captiveCoreBackend functions (breaking change in `ingest`)
* `PrepareCaptiveCore` needs the toml instead of configPath 
* addition of OperationTypes 14 -21


internal/input/assets.go
```diff
@@ -18,7 +19,8 @@ func GetPaymentOperations(start, end uint32, limit int64) ([]OperationTransformI
 
 	opSlice := []OperationTransformInput{}
 	for seq := start; seq <= end; seq++ {
-		txReader, err := ingest.NewLedgerTransactionReader(backend, publicPassword, seq)
+		ctx := context.Background()
```
sydneynotthecity, line 22: @acharb this prob needs to be moved up out of the for loop? wasn't sure on this

PR #129, opened 2021-08-06
Incorporate Protocol 17 Operations

sydneynotthecity:
**What**
With the release of Stellar Protocol 17, new concepts and subsequent operation types were introduced to the Network that were not accounted for in the `stellar-etl` SDK.  In particular:

* `Claimable Balances` - the ability to split up payments into parts that allows the sender to `CreateClaimableBalance` and the receiver to `ClaimClaimableBalance`. This added 2 new operations to the Network.
* `SponsoredReserves` - the ability to for an account to pay, or "sponsor", the base reserves of another account. This introduced 3 new operations: `BeginSponsoringFutureReserves`, `EndSponsoringFutureReserves` and `RevokeSponsorship`.
* `Clawbacks` - the ability for assets to burn their asset from a trustline or claimable balance. This adds 2 more operations to the network, `Clawback` and `ClawbackClaimableBalance`.
* `Set Trustlines` - the ability for an account to `SetTrustlineFlags` through a standalone operation instead of `AllowTrust` which has been deprecated.

**Why**
The DAGs that populate Hubble 2.0 tables are not reconciling with Hubble 1.0 because transaction and operation records that contain the new operations are getting dropped from the dataset. By adding the operations in the `transform/operations.go` file, the records will be written to the tables and no longer dropped.


internal/transform/schema.go
```diff
@@ -72,51 +76,68 @@ type OperationOutput struct {
...
-	StartingBalance    float64  `json:"starting_balance"`
-	To                 string   `json:"to"`
-	Trustee            string   `json:"trustee"`
-	Trustor            string   `json:"trustor"`
-	Value              string   `json:"value"` // base64 encoding of bytes for operations that manage data
-	ClearFlags         []int32  `json:"clear_flags"`
-	ClearFlagsString   []string `json:"clear_flags_s"`
-	DestinationMin     string   `json:"destination_min"`
-	BumpTo             string   `json:"bump_to"`
+	Account            string     `json:"account"`
```
acharb, line 79: ❓   what're your thoughts on refactoring the way we're doing this `Details` field. I'm not sure why we originally decided to do it this way, but to me it seems pretty unwieldy (both as Go code, and in big query). Could we break it up (would be in a different PR of course)? Or is there a reason the fields need to be grouped like this? (eg. other projects are dependent on the fields being at the same level in `.details` or something like that)

 By looking at it these are the added fields correct?
`AccountID`, `BalanceID`, `ClaimableBalanceID`, `Claimants`, `DataAccountID`, `DataName`, `SignerAccountID`, `SignerKey`, `TrustlineAccountID`, `TrustlineAsset`

nit: for PR readability if you could point out the added fields it would be helpful, took me a minute to compare to see which ones were added 🙏 😄

acharb (commented):
overall looks great! 💯 

I left some thoughts and questions inline

internal/transform/operation.go
```diff
@@ -94,6 +97,61 @@ func addAssetDetailsToOperationDetails(operationDetails *Details, asset xdr.Asse
...
+		operationDetails.SetFlagsString = s
+
+	case "clear":
+		operationDetails.ClearFlags = n
+		operationDetails.ClearFlagsString = s
+	}
+
+}
+
+func addLedgerKeyToDetails(operationDetails *Details, ledgerKey xdr.LedgerKey) error {
```
acharb, line 133: (1 more ❓) for my own knowledge what does ledgerKey represent?

internal/transform/operation.go
```diff
@@ -115,6 +173,28 @@ func transformPath(initialPath []xdr.Asset) []Path {
 	return path
 }
 
+func findInitatingBeginSponsoringOp(operation xdr.Operation, operationIndex int32, transaction ingest.LedgerTransaction) *SponsorshipOutput {
```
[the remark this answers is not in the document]
sydneynotthecity: I think this is a good idea. I have another task to update some of the unit tests and will include in that PR. We need to extend the `cmd` integration tests to test the new operations, i'll add the unit test at that time.

internal/transform/schema.go
sydneynotthecity: I dislike the current `Details` struct and had written a [new issue](https://github.com/stellar/hubble/issues/179) to restructure it when I fix the null handling. The way we currently structure `Details` does not allow for delineation between empty values and zero values, so the data does not match what is written to Horizon DB.

My hope was to refactor this to follow the Operations Protocol Design [here](https://github.com/stellar/go/blob/master/protocols/horizon/operations/main.go)... @acharb thoughts on us using this structure instead of our current one? Or were you thinking of flattening the `Details` object into a denormalized table in BQ?

sydneynotthecity: Other fields added: `BeginSponsor`, and `SponsoredID`

internal/transform/operation.go
sydneynotthecity: My understanding is that it is a struct comprising the `LedgerEntryType` with the unique identifier per `Type`.  Each state table in `stellar-core` has a unique ID per state table and the `LedgerKey` lets you access those IDs.

For example, if the `LedgerEntryType` is Account, you would be able to access the `AccountID`. Similarly, Claimable Balance would return the `BalanceID` etc

internal/transform/operation.go
```diff
@@ -13,10 +14,12 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type Claimants []Claimant
+
 // TransformOperation converts an operation from the history archive ingestion system into a form suitable for BigQuery
 func TransformOperation(operation xdr.Operation, operationIndex int32, transaction ingest.LedgerTransaction, ledgerSeq int32) (OperationOutput, error) {
 	outputTransactionID := toid.New(ledgerSeq, int32(transaction.Index), 0).ToInt64()
-	outputOperationID := toid.New(ledgerSeq, int32(transaction.Index), operationIndex).ToInt64()
+	outputOperationID := toid.New(ledgerSeq, int32(transaction.Index), operationIndex+1).ToInt64() //operationIndex needs +1 increment
```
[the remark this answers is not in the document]
sydneynotthecity: I do not have a great reason, I just noticed that our operation ids were off by 1 during data validation. Horizon increments by 1 [here](https://github.com/stellar/go/blob/ab0780781eece93b937043273bf090bf35dfc0f5/services/horizon/internal/ingest/processors/operations_processor.go#L92)....maybe just because the id is 1 indexed instead of 0 indexed?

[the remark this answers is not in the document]
acharb: ya taking a quick look at the `ingest` code I think you're right. Maybe comment could be `// operationIndex needs +1 increment to be in sync with ingest package`, or something like that 🤔

internal/transform/schema.go
acharb: resolving this since tracked in this ticket https://github.com/stellar/hubble/issues/179

internal/transform/operation.go
```diff
@@ -13,10 +14,12 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type Claimants []Claimant
```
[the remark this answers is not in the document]
sydneynotthecity: no reason, if we take out the type declaration, we'll just need to define the later variable as a slice before appending the `CreateClaimableBalance` details:
```
var claimants []Claimant
```
thoughts if this would be better?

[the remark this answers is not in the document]
acharb: yup I think so 👍

acharb (approved):
lgtm!

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
...
+	V0: &xdr.ClaimantV0{
+		Destination: testAccount1ID,
+		Predicate: xdr.ClaimPredicate{
+			Type: xdr.ClaimPredicateTypeClaimPredicateUnconditional,
+		},
+	},
+}
+
+var testClaimantDetails = Claimant{
+	Destination: testAccount1Address,
```
Eveeeehaleee, line 124: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
 var nativeAssetPath = Path{
 	AssetType: "native",
 }
+
+var genericClaimableBalance = xdr.ClaimableBalanceId{
```
Eveeeehaleee, line 108: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
...
+	Type: xdr.ClaimableBalanceIdTypeClaimableBalanceIdTypeV0,
+	V0:   &xdr.Hash{1, 2, 3, 4, 5, 6, 7, 8, 9},
+}
```
Eveeeehaleee, line 111: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
...
+
+var testClaimant = xdr.Claimant{
+	Type: xdr.ClaimantTypeClaimantTypeV0,
+	V0: &xdr.ClaimantV0{
```
Eveeeehaleee, line 115: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
 var nativeAssetPath = Path{
 	AssetType: "native",
 }
+
+var genericClaimableBalance = xdr.ClaimableBalanceId{
+	Type: xdr.ClaimableBalanceIdTypeClaimableBalanceIdTypeV0,
+	V0:   &xdr.Hash{1, 2, 3, 4, 5, 6, 7, 8, 9},
+}
+
+var testClaimant = xdr.Claimant{
```
Eveeeehaleee, line 113: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/test_variables_test.go
```diff
@@ -104,3 +104,23 @@ var nativeAsset = xdr.MustNewNativeAsset()
...
+	Type: xdr.ClaimantTypeClaimantTypeV0,
+	V0: &xdr.ClaimantV0{
+		Destination: testAccount1ID,
+		Predicate: xdr.ClaimPredicate{
```
Eveeeehaleee, line 117: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -44,7 +44,7 @@ func TestTransformOperation(t *testing.T) {
 		{
 			negativeOpTypeInput,
 			OperationOutput{},
-			fmt.Errorf("The operation type (-1) is negative for  operation 1 (operation id=4097)"),
+			fmt.Errorf("The operation type (-1) is negative for  operation 1 (operation id=4098)"),
 		},
 		{
```
Eveeeehaleee, line 49: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -101,6 +101,8 @@ func makeOperationTestInput() (inputTransaction ingest.LedgerTransaction, err er
 		Weight: xdr.Uint32(1),
 	}
```
Eveeeehaleee, line 102: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -101,6 +101,8 @@ func makeOperationTestInput() (inputTransaction ingest.LedgerTransaction, err er
...
 
+	hardCodedClaimableBalance := genericClaimableBalance
```
Eveeeehaleee, line 104: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -101,6 +101,8 @@ func makeOperationTestInput() (inputTransaction ingest.LedgerTransaction, err er
...
+	hardCodedClaimant := testClaimant
```
Eveeeehaleee, line 105: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -295,6 +297,158 @@ func makeOperationTestInput() (inputTransaction ingest.LedgerTransaction, err er
 				},
 			},
 		},
+		xdr.Operation{
+			SourceAccount: nil,
+			Body: xdr.OperationBody{
+				Type: xdr.OperationTypeCreateClaimableBalance,
+				CreateClaimableBalanceOp: &xdr.CreateClaimableBalanceOp{
```
Eveeeehaleee, line 304: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
```diff
@@ -295,6 +297,158 @@ func makeOperationTestInput() (inputTransaction ingest.LedgerTransaction, err er
 				},
 			},
 		},
+		xdr.Operation{
```
Eveeeehaleee, line 300: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation_test.go
Eveeeehaleee, line 304: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation.go
```diff
@@ -16,7 +17,7 @@ import (
 // TransformOperation converts an operation from the history archive ingestion system into a form suitable for BigQuery
 func TransformOperation(operation xdr.Operation, operationIndex int32, transaction ingest.LedgerTransaction, ledgerSeq int32) (OperationOutput, error) {
 	outputTransactionID := toid.New(ledgerSeq, int32(transaction.Index), 0).ToInt64()
-	outputOperationID := toid.New(ledgerSeq, int32(transaction.Index), operationIndex).ToInt64()
+	outputOperationID := toid.New(ledgerSeq, int32(transaction.Index), operationIndex+1).ToInt64() //operationIndex needs +1 increment to stay in sync with ingest package
 
 	outputSourceAccount, err := utils.GetAccountAddressFromMuxedAccount(getOperationSourceAccount(operation, transaction))
 	if err != nil {
```
Eveeeehaleee, line 23: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation.go
```diff
@@ -94,6 +95,61 @@ func addAssetDetailsToOperationDetails(operationDetails *Details, asset xdr.Asse
 	return nil
 }
```
Eveeeehaleee, line 96: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation.go
```diff
@@ -94,6 +95,61 @@ func addAssetDetailsToOperationDetails(operationDetails *Details, asset xdr.Asse
...
 
+func addTrustLineFlagToDetails(operationDetails *Details, f xdr.TrustLineFlags, prefix string) {
```
Eveeeehaleee, line 98: 0xa375D616941d1F4c489e225482ca4602df3178B2

internal/transform/operation.go
```diff
@@ -94,6 +95,61 @@ func addAssetDetailsToOperationDetails(operationDetails *Details, asset xdr.Asse
...
+	var (
+		n []int32
+		s []string
+	)
+
+	if f.IsAuthorized() {
```
Eveeeehaleee, line 104: 0xa375D616941d1F4c489e225482ca4602df3178B2

Eveeeehaleee (approved):

Issue #125, opened 2021-08-03
Add automated testing


acharb:
Add automated testing on new PRs similar to our other projects. This will ensure new code is tested before being able to merge in

shanzzam:
@acharb Is this an overall action item or a reminder? Do you intend to address testing this sprint or should this be on the backlog?

acharb:
it's a reminder, I'll get to it this sprint

PR #130, opened 2021-08-16
add github workflow for unit tests in internal package

acharb:
**WHAT**
adding automated unit testing for the `internal` package.

**WHY**
ensures code that's merged into master is tested.

closes https://github.com/stellar/stellar-etl/issues/125


.github/workflows/internal.yml
```diff
@@ -0,0 +1,33 @@
+name: Internal-unit-tests
```
sydneynotthecity, line 1: ❓ does the `name` have to match the yml file name or does that not matter?

.github/workflows/internal.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
+    - name: Build Internal
+      working-directory: internal
```
sydneynotthecity, line 19: is the idea that once we get the `cmd` tests working they would run as a separate workflow?

.github/workflows/internal.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1
+    - name: Run Internal Unit Tests
+      working-directory: internal
+      run: go test -v ./...
```
sydneynotthecity, line 33: 💡 it might be nice to add coverage with `-cover` in the future. I don't know if we'd like to keep a certain code coverage percentage in the future, but could help the reviewer see how well tested the code is.

sydneynotthecity (commented):
LGTM! 🎉 Just some questions for my understanding and one idea

sydneynotthecity (approved):

.github/workflows/internal.yml
acharb: yup 👍 . Or we can change it to do a global `gotest -v ./...` to run all the unit tests in the same job, once the `cmd` ones are fixed

.github/workflows/internal.yml
acharb: technically it doesn't, but for consistency makes sense to me to keep it the same, changed here:
https://github.com/stellar/stellar-etl/pull/130/commits/47bbd8c40ad6c0c60720b4d826c25d571da4a94a

.github/workflows/internal.yml
acharb: ooh TIL about `-cover` 🙌 . I think that's a good idea added
https://github.com/stellar/stellar-etl/pull/130/commits/d27f7fb8786f7bc758f7b447307aebc4d18938e3

acharb:
btw terraform PR for blocking merges if these tests fail here:
https://github.com/stellar/terraform/pull/504

PR #133, opened 2021-08-27
Add Fee Bumps, Muxed Accounts to Transactions

sydneynotthecity:
Two important concepts were left off the `transform/transactions.go` file that need to be introduced to Hubble 2.0: 
1. Fee Bumps
2. Muxed Accounts

This PR introduces both to Hubble 2.0. Changes were minor, focusing primarily on

#### Schema Changes
The following fields were added to the `TransactionOutput` struct:

* `AccountMuxed`
* `FeeAccount`
* `FeeAccountMuxed`
* `InnerTransactionHash`
* `NewMaxFee`

#### New Code
* Code to parse InnerTransactionResult when there is a `FeeBump` included in the transaction
* Code to parse `MuxedAccounts` if the crypto signature matches

#### Unit Tests
An additional unit test was written for `transform/transactions.go`, which tests a successful Fee Bumped transaction. In order to add a unit test, the `makeTransactionTestInput` and `makeTransactionTestOutput` functions were refactored. In prior versions, both could only support structs. The type was updated to be an _array_ of structs for ease of multiple tests during test execution.

### DIFF
--- a/internal/transform/schema.go
+++ b/internal/transform/schema.go
@@ -33,6 +33,7 @@ type TransactionOutput struct {
 	LedgerSequence       uint32    `json:"ledger_sequence"`
 	ApplicationOrder     uint32    `json:"application_order"`
 	Account              string    `json:"account"`
+	AccountMuxed         string    `json:"account_muxed,omitempty"`
 	AccountSequence      int64     `json:"account_sequence"`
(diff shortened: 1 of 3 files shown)


nikhilsaraf (approved):
Lgtm, thanks!

PR #134, opened 2021-09-08
Add Logging, Fix Transactions Bug

sydneynotthecity:
Current `stellar-etl` cmd options do not provide robust logging to the end user when exporting data from history archives. The current `cmd/export_*` scripts will print processing stats to the screen, however the print statements do not carry through to other systems (like Airflow). All print statements were converted to `log.INFO` statements so that any Loggers with the log level set to `INFO` will log the processing stats. 

One additional bug was caught and fixed. In the `transactions` transformation, valid transactions were thrown out due to a faulty check. There is an error condition that checks that max > min in the `time_bounds` field. However, if there is no upper bound, `stellar-core` sets the [boundary to empty](https://github.com/stellar/stellar-core/blob/dd9ab5769d12a0bba0404d65365e468f8fdaa379/src/xdr/Stellar-transaction.x#L576), which is 0. The error condition was updated to require max > 0 in order to fail.

#### Logging Changes
* All `cmd_*` loggers are set to log level of `INFO`
* All `cmd_*` scripts now log the transformation stats instead of printing them
* `cmd_*` scripts will log an error if it has trouble writing to a file. Previously the errors were swallowed.
* A new processing stat is introduced: `Bytes Written`. If the user writes the exported data to a file, the total number of bytes written is logged. This helps with debugging when if the user suspects a partial file has been written. 

#### Transaction Changes
* Updated the time bounds check to have two conditions: max > min AND max > 0 since empty max time bounds are set to 0. If the max time == 0, the end boundary condition is not included
* The `time_bounds` format deletes the `' '` to conform with Horizon. ie `[0, 100)` becomes `[0,100)`
* Unit tests are updated to pass with new conditions


cmd/export_accounts.go
```diff
@@ -60,15 +63,25 @@ the export_ledger_entry_changes command.`,
 			}
 
 			if !useStdout {
-				outFile.Write(marshalled)
+				nb, err := outFile.Write(marshalled)
+				numBytes += nb
 				outFile.WriteString("\n")
+				if err != nil {
+					cmdLogger.Info("Error writing accounts to file: ", err)
+				}
```
acharb, line 71: 💡  moving the `if err != nil` line up. Generally that check comes right after the line the `err` is returned. I don't think it matters much here for functionality but helps for readability
```suggestion
				nb, err := outFile.Write(marshalled)
				if err != nil {
					cmdLogger.Info("Error writing accounts to file: ", err)
				}
				numBytes += nb
				outFile.WriteString("\n")
				
```

acharb: I think there are a couple spots below that are similar

cmd/export_accounts.go
```diff
@@ -36,6 +38,7 @@ the export_ledger_entry_changes command.`,
 		}
 
 		failures := 0
+		numBytes := 0
```
acharb, line 41: ❓ jw is numBytes the most helpful way to see progress? Or would `numOfAccounts` or something like that be more helpful. bytes is less clear to me but up to you

cmd/export_operations.go
```diff
@@ -59,15 +63,26 @@ var operationsCmd = &cobra.Command{
 			}
 
 			if !useStdout {
-				outFile.Write(marshalled)
+				// outFile.Write(marshalled)
```
acharb, line 66: ❓ is this comment left on purpose?

internal/transform/transaction.go
```diff
@@ -69,12 +69,18 @@ func TransformTransaction(transaction ingest.LedgerTransaction, lhe xdr.LedgerHe
 	timeBound := transaction.Envelope.TimeBounds()
 	outputTimeBounds := ""
 	if timeBound != nil {
-		if timeBound.MaxTime < timeBound.MinTime {
+		if timeBound.MaxTime < timeBound.MinTime && timeBound.MaxTime > 0 {
```
acharb, line 72: 💡 not sure if this matters but the case of timeBound.MaxTime being negative value isn't caught. Not sure if that's a possible value, but just to be safe this line could be:

```suggestion
		if timeBound.MaxTime < timeBound.MinTime && timeBound.MaxTime != 0 {
```

alternatively could make the line a couple lines down:
`if timeBound.MaxTime <= 0 {`

acharb (commented):
looks great 🙌  left a couple comments inline

internal/transform/transaction.go
acharb: re-reading your description it doesn't look negative is a possible value, so you can ignore this comment if you don't think the change is needed

cmd/export_accounts.go
sydneynotthecity: thanks, i'll make the necessary updates to each `cmd_*` script

cmd/export_operations.go
sydneynotthecity: No, good catch. Deleting now

internal/transform/transaction.go
sydneynotthecity: Committing the suggestion just in case there is an exception in the future

cmd/export_accounts.go
sydneynotthecity: The [printTransformStats](https://github.com/stellar/stellar-etl/blob/master/cmd/export_ledgers.go#L51-L54) function already reports the number of "transforms" (in this case, should be accounts) because it is the number of unique records in the file. 

I added `numBytes` so that we know how large the file is that is written to local storage. The reason I did this is the following Airflow task reports how large the file is that is uploaded to GCS. When debugging, we can compare the reported file sizes as a check to see if the file we uploaded to GCS is the same size as the file written to the local volume in the Docker container.

acharb (approved):
lgtm!

PR #144, opened 2021-11-02
Add --gcs-bucket flag that indicates a location on GCS to upload exported data to

erika-sdf:
Specifying `--gcs-bucket` (or alternatively `-g`) will upload the resulting output file into the specified bucket.

For example, the following command:
```
stellar-etl export_ledgers -s 100000 -e 100001 -g stellar-etl-cli -o testdata/
```
will produce the a file containing the output at 
```
gs://stellar-etl-cli/testdata/ledgers-100000-100001.txt
```

For local runs, credentials can be provided by passing the path of a service account json to `--gcp-credentials`. In production, the `gcp-credentials` flag should be omitted, and credentials should be pulled from the environment's default service account.


cmd/export_liquidity_pools.go
```diff
@@ -24,67 +22,46 @@ should be used in an initial data dump. In order to get liqudity pools informati
...
 			if err != nil {
 				if strictExport {
 					cmdLogger.Fatal("could not transform pool", err)
 				} else {
 					cmdLogger.Warning("could not transform pool", err)
-					failures++
+					numFailures++
 					continue
 				}
 			}
```
sydneynotthecity, line 50: Was this missed in refactor? Does not match the rest of the `cmd` utility files

cmd/export_accounts.go
```diff
@@ -24,79 +22,56 @@ should be used in an initial data dump. In order to get account information with
 the export_ledger_entry_changes command.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		cmdLogger.SetLevel(logrus.InfoLevel)
-		endNum, useStdout, strictExport, isTest := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
+		endNum, strictExport, isTest := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
+		cmdLogger.StrictExport = strictExport
```
sydneynotthecity, line 26: Do we ever do anything with the `StrictExport` flag? Is there merit in us even keeping this bool in the code? Afaik I can't see the utility for it.

cmd/export_ledgers.go
```diff
@@ -54,13 +82,56 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
-	if printLog {
-		fmt.Println(string(results))
-	} else {
-		cmdLogger.Info(string(results))
+func maybeUpload(gcpCredentials, gcsBucket, runId, path string) {
+	if len(gcsBucket) > 0 {
+		err := uploadToGcs(gcpCredentials, gcsBucket, runId, path)
+		if err != nil {
+			cmdLogger.Errorf("Unable to upload output to GCS: %s", err)
+		}
```
sydneynotthecity, line 134: 💡 idea: would it be helpful to add an upload success logging statement as well? Let the user know that the file was successfully uploaded, perhaps with number of bytes transferred?

cmd/export_ledgers.go
```diff
@@ -54,13 +82,56 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
+		return fmt.Errorf("failed to create client: %v", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*60)
+	defer cancel()
+
+	objectPath := filepath.Join(runId, path)
+	wc := client.Bucket(bucket).Object(objectPath).NewWriter(ctx)
+	defer wc.Close()
```
sydneynotthecity, line 119: Would it be useful to check the `err` in the `.Close()` method? Are there cases in GCS where the `.Close()` function would error and we wouldn't know? (Like timeouts or incomplete file writes?)

cmd/export_ledgers.go
```diff
@@ -54,13 +82,56 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
+
+	cmdLogger.Infof("Uploading %s to gs://%s/%s ", path, bucket, objectPath)
+
+	if _, err := io.Copy(wc, reader); err != nil {
```
sydneynotthecity, line 123: Are there any limitations to file size that can be uploaded? Sometimes with historical loading the `operations` cmd utility can generate 1-2GB files.

cmd/export_ledgers.go
```diff
@@ -54,13 +82,56 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
+	}
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create client: %v", err)
+	}
+	defer client.Close()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*60)
```
sydneynotthecity, line 114: ❓ does this timeout apply to any operation executed by the client?

cmd/export_assets.go
```diff
@@ -18,86 +14,60 @@ var assetsCmd = &cobra.Command{
...
-				errMsg := fmt.Sprintf("could not extract asset from operation %d in transaction %d in ledger %d: ", transformInput.OperationIndex, txIndex, transformInput.LedgerSeqNum)
-				if strictExport {
-					cmdLogger.Fatal(errMsg, err)
-				} else {
-					cmdLogger.Warning(errMsg, err)
-					failures++
-					continue
-				}
-
+				cmdLogger.Errorf("could not extract asset from operation %d in transaction %d in ledger %d: ", transformInput.OperationIndex, txIndex, transformInput.LedgerSeqNum)
```
sydneynotthecity, line 37: FYI the other command utilities use `LogErrror()` instead of `Errorf()` here

cmd/get_ledger_range_from_times.go
```diff
@@ -77,9 +66,11 @@ var getLedgerRangeFromTimesCmd = &cobra.Command{
 			cmdLogger.Fatal("could not json encode ledger range", err)
 		}
 
-		if !useStdout {
+		if path != "" {
+			outFile := mustOutFile(path)
```
sydneynotthecity, line 70: This is clever, I like this implementation :)

internal/utils/main.go
```diff
@@ -186,8 +185,7 @@ func AddLPOperations(txMeta []xdr.OperationMeta, AssetA, AssetB xdr.Asset) []xdr
 // AddCommonFlags adds the flags common to all commands: end-ledger, stdout, and strict-export
 func AddCommonFlags(flags *pflag.FlagSet) {
 	flags.Uint32P("end-ledger", "e", 0, "The ledger sequence number for the end of the export range")
-	flags.Bool("stdout", false, "If set, the output will be printed to stdout instead of to a file")
-	flags.Bool("strict-export", false, "If set, transform errors will be reported as fatal errors instead of warnings.")
+	flags.Bool("strict-export", false, "If set, transform errors will be fatal.")
```
sydneynotthecity, line 188: Disregard `StrictExport` comment above, I see the use

internal/utils/main.go
```diff
@@ -278,8 +278,22 @@ func MustBucketFlags(flags *pflag.FlagSet, logger *log.Entry) (path string) {
 	return
 }
 
+// MustGcsFlags gets the values of the bucket list specific flags: gcp-project and gcs-bucket
+func MustGcsFlags(flags *pflag.FlagSet, logger *EtlLogger) (bucket, credentials string) {
+	bucket, err := flags.GetString("gcs-bucket")
+	if err != nil {
+		logger.Fatal("could not get gcs bucket: ", err)
+	}
+	credentials, err = flags.GetString("gcp-credentials")
```
sydneynotthecity, line 287: ❓ Does this mean we still need to pass the gcp-credentials location when it executes out of cloud composer as well? With the assumption we'll pass the default credentials?

sydneynotthecity (commented):
Great work! Had a couple comments for my own understanding, but overall looks really good. Thanks for all your work on the refactor; the code is much improved 🚀

cmd/export_liquidity_pools.go
erika-sdf: Thanks for catching this. I have changed it to look like the rest.

cmd/export_accounts.go
erika-sdf: I think what we get out of it is to choose what the tool does when an entry during export fails. The tool crashes with 'strict_export'. Without it, it logs an error and continues onto the next item instead. We should probably decide whether there's a reason to make this behavior customizable, or if only one of the behaviors are needed.

I think we want `strict_export` behavior (failing loudly) in prod. I can imagine allowing failures to pass through maybe being useful in dev, but unsure if that will really be the case. What do you think?

cmd/export_ledgers.go
erika-sdf: Sure. I have added one to `uploadToGcs`

cmd/export_ledgers.go
erika-sdf: Agreed that it would be helpful to surface that error. Added error checking for it.

cmd/export_ledgers.go
erika-sdf: Seems like the limit for one object is [5 TB](https://cloud.google.com/storage-transfer/docs/known-limitations-transfer), so 1-2GB should theoretically be ok.

cmd/export_assets.go
erika-sdf: Good catch. Updated.

cmd/get_ledger_range_from_times.go
erika-sdf: :)

internal/utils/main.go
erika-sdf: In the cloud composer environment, the credentials should be assumed by the default service account, so this flag should not be specified.

internal/utils/main.go
erika-sdf: Acknowledged. Will leave it in for now, unless there's more discussion on the topic

cmd/export_accounts.go
sydneynotthecity: I agree. That is useful and should be set in the production env

cmd/export_ledgers.go
```diff
@@ -54,13 +82,62 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
-	} else {
-		cmdLogger.Info(string(results))
+	cmdLogger.Info(string(results))
+}
+
+func generateRunId() string {
+	return uuid.New().String()
+}
+
+func uploadToGcs(credentialsPath, bucket, runId, path string) error {
```
sydneynotthecity, line 95: One other idea: does it make sense to include the `job_name` in the GCS URI? Curious if you think it would make debugging easier if ever we needed to look at the data. 
Instead of
```
gs://stellar-etl-cli/<run_id>/<filename>
```
Add a directory by `job_name` (using `export-accounts` as example):
```
gs://stellar-etl-cli/export-accounts/<run_id>/<filename>
```

cmd/export_ledgers.go
```diff
@@ -54,13 +82,62 @@ func printTransformStats(attempts, failures int, printLog bool) {
...
+	objectPath := filepath.Join(runId, path)
+	wc := client.Bucket(bucket).Object(objectPath).NewWriter(ctx)
+
+	cmdLogger.Infof("Uploading %s to gs://%s/%s", path, bucket, objectPath)
+
+	var written int64
+	if written, err = io.Copy(wc, reader); err != nil {
+		return fmt.Errorf("unable to copy: %v", err)
+	}
+	err = wc.Close()
```
sydneynotthecity, line 126: Do we plan on deleting the local file if GCS upload is successful? Or would you rather the crawler in the `airflow-workers` remove files after `X` minutes have passed?

sydneynotthecity (approved):
Had two additional thoughts after testing locally. Optional changes that could be helpful

cmd/export_ledgers.go
erika-sdf: I was also thinking more about the output file structure after looking at how it is used in the dags. What do you think of getting rid of run_id and doing something like: 

```
gs://stellar-etl-cli/export-accounts/<filename>
```

or 

```
gs://stellar-etl-cli/accounts/<filename>
```

sydneynotthecity: I'm in favor of that. I think it's simpler and makes sense. With that structure would the file names be unique per executed batch? Like:
```
accounts-<run_id>.txt
```
or 
```
<ledger-range>.txt
```
etc

cmd/export_ledgers.go
erika-sdf: I don't mind adding it, but is there an urgency to have the files deleted? Are we worried about running out of disk for unbounded runs?

sydneynotthecity: Not for the unbounded runs, but for historical loads with parallelized execution we could quickly run out of disk space if we do not spec the `airflow-workers` appropriately (set to 32GB of shared storage rn). I'm ok if we include the feature in a later PR or keep the existing `poststart.sh` script on the `airflow-workers`. That `poststart.sh` script today will clean up the old files, but in general I would like old files cleaned up to keep persisted storage small. 

Thoughts on keeping the current design, a long-running script that does the cleanup in the background of the `airflow-workers`, versus adding local file deletion to the `stellar-etl` image?

cmd/export_ledgers.go
erika-sdf: After giving this some thought, I decided to push this problem up a layer, and let the airflow code decide how to namespace the data from each run. The resulting behavior is that the `stellar-etl` CLI naively takes the output path provided, and writes the content there. 

So, invoking 

```
stellar-etl export_ledgers -s 100000 -e 100001 -g stellar-etl-cli -o testdata/<run_id>
```

will produce 

```
testdata/<run_id>/100000-10001-accounts.txt
```

which will then be uploaded to

```
gs://stellar-etl-cli/testdata/<run_id>/testdata/<run_id>/100000-10001-accounts.txt
```

The local copy is deleted after that.

cmd/export_ledgers.go
erika-sdf: Seems cleaner to just do it here. Added some code that cleans up files after they are uploaded to GCS. Exported files that are not uploaded are left alone.

erika-sdf:
This now works with the `export_bounded_core` dag, which I will send you in a follow up CL.

internal/transform/schema.go
```diff
@@ -149,7 +149,7 @@ type TrustlineOutput struct {
 	AssetIssuer        string      `json:"asset_issuer"`
 	AssetType          int32       `json:"asset_type"`
 	Balance            float64     `json:"balance"`
-	TrustlineLimit     int64       `json:"trust_line_limit"`
+	TrustlineLimit     int64       `json:"trustline_limit"`
```
sydneynotthecity, line 152: The `TrustlineLimit` column in the `trust_lines` table is still `trust_line_limit`. Is there a reason that we're changing the schema to `trustline_limit` here? Just want to make sure they continue matching.

cmd/export_orderbooks.go
```diff
@@ -89,42 +88,76 @@ var exportOrderbooksCmd = &cobra.Command{
...
+			if err != nil {
+				return err
+			}
+			for k, v := range extra {
+				i[k] = v
+			}
+			bytesToWrite, err = json.Marshal(i)
+			if err != nil {
+				return err
+			}
```
sydneynotthecity, line 115: ❓ if the `export_orderbooks` is processing multiple batches, does this mean that there could be multiple lines of run metadata (`extra`) written to the same file? Ie would there be a metadata record _per_ batch in the file?

sydneynotthecity (approved):
Just had a question for my understanding, otherwise LGTM!

cmd/export_orderbooks.go
erika-sdf: The extra fields are appended to every single row in the output, since they're associated with every run in BigQuery. It is not per batch, but per row.

sydneynotthecity: Tracking now. I has assumed this was the `{transforms: 20, failed_transforms:0, etc}` record. Makes sense

internal/transform/schema.go
erika-sdf: You're right that this shouldn't be included. Removed.

PR #146, opened 2021-11-17
Set UseNumber for json decoder for max ints are stored properly

erika-sdf:


sydneynotthecity (approved):
LGTM! Thanks for getting that fix in :)

PR #147, opened 2021-11-22
Remove check for trades with 0 selling and buying amounts.

erika-sdf:
This is causing spammy logs/failures for `export_trade` jobs. `stellar-core` garbage collects trades with 0 values.


sydneynotthecity (approved):
LGTM!

PR #149, opened 2021-12-20
Comment out live core 003 from Archive URLs.

erika-sdf:
This is temporary. `core_live_003` seems to be refusing requests, and all runs are failing, which seems unexpected.


sydneynotthecity (approved):
LGTM

PR #151, opened 2022-01-11
Add account signers to Hubble 2.0

erika-sdf:


cmd/export_ledger_entry_changes.go
```diff
@@ -99,9 +102,18 @@ be exported.`,
 							if err != nil {
 								entry, _, _, _ := utils.ExtractEntryFromChange(change)
 								cmdLogger.LogError(fmt.Errorf("error transforming account entry last updated at %d: %s", entry.LastModifiedLedgerSeq, err))
+							} else {
+								transformedOutputs["accounts"] = append(transformedOutputs["accounts"], acc)
+							}
```
sydneynotthecity, line 107: ❓  Should this be wrapped in an `else` statement? The rest of the `LedgerEntryType` append the `LedgerType` regardless if an error is returned.

sydneynotthecity (commented):
1 question, and looks like a couple unit tests are failing. Once those pass, i'll approve

cmd/export_ledger_entry_changes.go
erika-sdf: I think in other instances where there is an error during `Transform*`, the error check hits `continue`, and the next `ledgerEntry` is considered. 

In this case, both `signers` and `account` depend on the same `ledgerEntry`. I guess it depends on whether we still want to try to extract signers if `TransformAccount` returns an error. I think it's reasonable to just bail as well.

erika-sdf:
This is ready for review again

internal/transform/account_signer.go
```diff
@@ -0,0 +1,43 @@
...
+	ledgerEntry, changeType, outputDeleted, err := utils.ExtractEntryFromChange(ledgerChange)
+	if err != nil {
+		return signers, err
+	}
+	outputLastModifiedLedger := uint32(ledgerEntry.LastModifiedLedgerSeq)
+	accountEntry, accountFound := ledgerEntry.Data.GetAccount()
+	if !accountFound {
+		return signers, fmt.Errorf("could not extract signer data from ledger entry of type: %+v", ledgerEntry.Data.Type)
+	}
+	sponsors := accountEntry.SponsorPerSigner()
```
sydneynotthecity, line 24: Can there be multiple sponsors on an account? And if so, how are we parsing each sponsor individually?

cmd/export_ledger_entry_changes.go
```diff
@@ -101,7 +104,18 @@ be exported.`,
 								cmdLogger.LogError(fmt.Errorf("error transforming account entry last updated at %d: %s", entry.LastModifiedLedgerSeq, err))
 								continue
 							}
-							transformedAccounts = append(transformedAccounts, acc)
+							transformedOutputs["accounts"] = append(transformedOutputs["accounts"], acc)
+							if change.AccountSignersChanged() {
```
sydneynotthecity, line 108: If the accounts signers change and that is the only change to the account, does the `accounts` data change at all? Or would we get a "changed" record that duplicated the last ledger modified sequence?

sydneynotthecity (approved):
Couple questions otherwise LGTM

internal/transform/account_signer.go
erika-sdf: I think there can be up to one sponsor per signer, and multiple signers per account.  The sponser per signer is extracted in the loop on L27.

cmd/export_ledger_entry_changes.go
erika-sdf: I see. I added a conditional to only update `accounts` only if `AccountChangedExceptSigners`.

PR #154, opened 2022-03-17
Add sponsorship to offers and accounts

sydneynotthecity:
Sponsorship was missing from two state tables. The following missing fields were added: 

**`accounts`**
* `sponsor`
* `num_sponsored`
* `num_sponsoring`

**`offers`**
* `sponsor`

Associated unit tests were updated for both to include `sponsor` in the `LedgerEntry` XDR.

Additionally, I decided to change the schema of the `offers` table to make the user experience better. Today, the buying and selling assets are reported by their id, forcing you to join to the `history_assets` table. I decided to normalize those fields:
`buying_asset_id` --> updated to include asset_type, asset_code and asset_issuer
`selling_asset_id` --> updated to include asset_type, asset_code and asset_issuer

This makes the table more readable and reduces the number of joins necessary.


erika-sdf (approved):
LGTM

PR #155, opened 2022-03-23
Add dust trades

sydneynotthecity:
### Overview

Liquidity Pools has a rounding mechanism that causes trade prices to slip when trading in small amounts. 
Since the final traded amount is the only reported amount, we would like to know when a trade amount has slipped.

### Changes

* Add `TradeType` int32, an ordinal field where 1 = orderbook trades and 2 = liquidity pool trades
* Add `RoundingSlippage` null.Int that represents the slippage amount of liquidity pool trades, in basis points
* Add `SellerIsExact` a boolean that represents if the seller amount is the exact representation of the amount (or if it is the slipped amount)
* Update unit tests to add new fields

Closes [GH#258](https://github.com/stellar/hubble/issues/258)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17.5
+FROM docker.io/golang:1.18
 
 COPY . /etl
 WORKDIR /etl
--- a/go.mod
+++ b/go.mod
@@ -1,39 +1,81 @@
 module github.com/stellar/stellar-etl
 
-go 1.16
+go 1.17
 
 require (
 	cloud.google.com/go/storage v1.10.0
+	github.com/guregu/null v2.1.3-0.20151024101046-79c5bd36b615+incompatible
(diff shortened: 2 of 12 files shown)


erika-sdf (commented):
Looks good overall. Is it possible to add a test case for `tradeType = 2` and one where `roundingSlippage` is not empty?

sydneynotthecity:
Yeah, I think that's a good idea. Let me see how hard that would be to add.

sydneynotthecity:
I've added two new unit test cases to `trades`:
1. Liquidity pool trade that is a `PathPaymentStrictSend`. The rounding slippage is 0 and validates `TradeType`
2. Liquidity pool trade that is a `PathPaymentStrictReceive` of a super small amount. Rounding slippage is 100 due to trade size.

erika-sdf (approved):

erika-sdf (approved):
Thanks for adding more complete test coverage. Looks good!

PR #156, opened 2022-04-04
Protocol 19

sydneynotthecity:
Updates to ETL primarily include new column additions to the `accounts` and `history_transactions` table

**`accounts`**

* **add** `sequence_time` type `INTEGER`

**`history_transactions`**

* **add** `ledger_bounds` string which contains min and max ledger bounds when not null (else empty string)
* **add** `min_account_sequence` type `NULL INTEGER`
* **add** `min_account_sequence_age` type `NULL INTEGER`
* **add** `min_account_sequence_ledger_gap` type `NULL INTEGER`
* **add** `min_account_sequence_ledger_gap` type `NULL INTEGER`
* **add** `extra_signers` type `REPEATED FIELD` which is a string array

**TODO**

- [x] Update `horizon` and `ingest` packages to protocol 19 version once deployed to testnet
- [x] Update `claimants` to include transaction conditions?
- [x] Add test cases for transactions tests
- [x] Change docker image in test to use new changes and observe for 1 month

Closes [GH#253](https://github.com/stellar/hubble/issues/253)


sydneynotthecity:
The `claimant` XDR was not updated. No action necessary.

internal/transform/schema.go
```diff
@@ -30,25 +32,30 @@ type LedgerOutput struct {
...
+	TimeBounds                  string         `json:"time_bounds"`
+	Successful                  bool           `json:"successful"`
+	TransactionID               int64          `json:"id"`
+	FeeAccount                  string         `json:"fee_account,omitempty"`
+	FeeAccountMuxed             string         `json:"fee_account_muxed,omitempty"`
+	InnerTransactionHash        string         `json:"inner_transaction_hash,omitempty"`
+	NewMaxFee                   uint32         `json:"new_max_fee,omitempty"`
+	LedgerBounds                string         `json:"ledger_bounds"`
+	MinAccountSequence          null.Int       `json:"min_account_sequence"`
+	MinAccountSequenceAge       null.Int       `json:"min_account_sequence_age"`
```
sreuland, line 56: in the[ xdr model on new p19 preconditions](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0021.md#specification), `MinAccountSequence` was nullable, but the other two(SequenceAge/LedgerGap) were non-nullable, seems like could be safely typed a bit more strongly here to same, but maybe this is already more into the extraction model now, just wanted to mention.

internal/transform/transaction_test.go
```diff
@@ -134,9 +134,11 @@ func makeTransactionTestInput() (transaction []ingest.LedgerTransaction, history
 							Text: &hardCodedMemoText,
 						},
 						Fee: 90000,
-						TimeBounds: &xdr.TimeBounds{
-							MinTime: 0,
-							MaxTime: 1594272628,
+						Cond: xdr.Preconditions{
```
sreuland, line 137: minor suggestion, could define some `ExtraSigners` to trigger coverage over that formatting routine.

sreuland (approved):
looks good, tests pass, I haven't worked in the repo before, left a few comments for consideration, my approval may not be the best endorsement!

internal/transform/transaction_test.go
sydneynotthecity: good point, i'll incorporate another test that validates more of the Protocol 19 preconditions

internal/transform/schema.go
sydneynotthecity: @sreuland do you know why [Horizon](https://github.com/stellar/go/blob/6af10a1f8a3f6d6d5775d8d2718779361f3c1b6d/services/horizon/internal/db2/history/transaction_batch_insert_builder.go#L136-L138) defines these preconditions as `null.Int`? For most data types in Hubble, I try and mirror whatever is implemented in Horizon DB. (Maybe @Shaptic knows?)

I think your hypothesis is correct, I just want to confirm.

sreuland: I see, followed through horizon references a bit on these attributes, and yes, they use nullable references, probably because prior to p19, tx's wouldn't have presence of that yet, the usage here is consistent, sorry for the detour, looks fine!

PR #157, opened 2022-04-29
Fix trades bug

sydneynotthecity:
Closes bug [GH260](https://github.com/stellar/hubble/issues/260)


internal/transform/trade.go
```diff
@@ -106,9 +106,9 @@ func TransformTrade(operationIndex int32, operationID int64, transaction ingest.
 		}
 
 		if BuyingOffer != nil {
-			outputBuyingOfferID = null.IntFrom(int64(claimOffer.OfferId()))
+			outputBuyingOfferID = null.IntFrom(int64(BuyingOffer.OfferId))
```
Shaptic, line 109: Copy-paste strikes again eh?

internal/transform/trade.go
```diff
@@ -106,9 +106,9 @@ func TransformTrade(operationIndex int32, operationID int64, transaction ingest.
...
 		} else {
-			outputBuyingOfferID = null.IntFrom(toid.EncodeOfferId(uint64(operationID), toid.TOIDType))
+			outputBuyingOfferID = null.IntFrom(toid.EncodeOfferId(uint64(operationID)+1, toid.TOIDType))
```
Shaptic, line 111: Why +1? Maybe worth a `// comment`

Shaptic (approved):
I don't really know what I'm reviewing but the code LGTM besides one question :arrow_down:

erika-sdf (approved):

internal/transform/trade.go
sydneynotthecity: In the worst possible way

internal/transform/trade.go
sydneynotthecity: it's a quirk with the `ingest` package where the `operation_id` is off by 1. I'll add a comment

PR #158, opened 2022-05-16
Fix claimable balance id 

sydneynotthecity:
Closes [#272](https://github.com/stellar/hubble/issues/272)

The `history_operations` and `claimable_balances` table report the `balance_id` differently. The `history_operations` table includes `"00000000"` to the front of the id, which is the correct way to report the id. Balance IDs were truncated  in `claimable_balances` due using the `GetV0()` method. Transformation is corrected so that both align.

### DIFF
--- a/internal/transform/claimable_balance.go
+++ b/internal/transform/claimable_balance.go
@@ -2,6 +2,7 @@ package transform
 
 import (
 	"fmt"
+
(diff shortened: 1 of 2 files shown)


erika-sdf (approved):
LGTM

PR #159, opened 2022-05-20
Capture all state changes

sydneynotthecity:
State tables were only capturing changes to state occurring in the _last_ ledger within block range. This was due to us assigning the `changeCompactor` the last set of changes (last ledger) instead of _appending_ the change set. 

PR also adds `--export-balances` as an option from command line to be used when debugging.


erika-sdf (approved):
LGTM

PR #160, opened 2022-05-25
Remove duplicate processed ledgers from state tables

sydneynotthecity:
The state table processors in Hubble were introducing duplicate records to the tables for two reasons: 
1. Given a batch range, the `export_ledger_entry_changes` code incremented the `end-ledger` by 1, which introduced a 1-ledger overlap between end ledger values and start ledger values for the following batch. 
2. The `input/changes.go` code divided up large batches into checkpoint-sized batches. These checkpoint sized definitions processed ledger inclusive of both start/end ledger ranges. This would double process end ledgers as the following batch's start ledger.

Changes adjust the processors to be exclusive of end ledgers so that the range processes each ledger only once.

Closes [#271](https://github.com/stellar/hubble/issues/271)

### DIFF
--- a/cmd/export_ledger_entry_changes.go
+++ b/cmd/export_ledger_entry_changes.go
@@ -62,9 +62,6 @@ be exported.`,
 			cmdLogger.Fatal("could not get absolute filepath for the config file: ", err)
 		}
 
-		if endNum != 0 {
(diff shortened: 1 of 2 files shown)


erika-sdf (approved):
LGTM

PR #162, opened 2022-05-27
Adjust batch starts and filenames

sydneynotthecity:
The batch start needed to be incremented by 1 so that ledger entry ranges did not overlap. The overlapping was still causing duplicates in the final dataset.

Adjustments also made to the filenames since the `exportFilename` method decrements the ledger range by 1. Incremented the end input by 1 so that we preserve the correct range in the filename.


sreuland (approved):
lgtm, nice test verification.

PR #167, opened 2022-08-30
Fix PATH and remove GOPATH

tamirms:
This fixes the build failures in https://buildmeister-v3.stellar-ops.com/job/ProductEngineering/job/docker-stellar-etl/25/execution/node/24/log/


Shaptic (approved):
There's no chance the order change matters, right? In that the `wget`, etc. come before the `ENV` stuff? My gut says no but just sanity checking.

tamirms:
> There's no chance the order change matters, right? In that the wget, etc. come before the ENV stuff? My gut says no but just sanity checking.


Correct, the environment variable changes are not required for installing go

PR #168, opened 2022-09-16
Export history effects

lucaszanotelli:
**Overview**

This change adds effects processors from Horizon to `stellar-etl`.
This also adds support to `offer_created`, `offer_updated`, `offer_removed` effects.

**Changes**

- add effects related schema, transformation and exporting files
- update `README` to include latest modifications
- include functions to `transform/operation.go` to support `transform/effect.go`
- add related test data and test files

Partially closes stellar/hubble [#246](https://github.com/stellar/hubble/issues/246)
Closes stellar/go [#166](https://github.com/stellar/go/issues/166)

Obs.: most of the work related to transformation was adapting code from Horizon processors to `stellar-etl`.
Here's the source file and the result:

horizon (source) => stellar-etl (result)
[`effects_processor.go`](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/effects_processor.go) => [`transform/effects.go`](https://github.com/stellar/stellar-etl/blob/export_history_effects/internal/transform/effects.go)
[`effects_processor_test.go`](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/effects_processor_test.go) => [`transform/effects_test.go`](https://github.com/stellar/stellar-etl/blob/export_history_effects/internal/transform/effects_test.go)
[`operations_processor.go`](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/operations_processor.go) => [`transform/operations.go`](https://github.com/stellar/stellar-etl/blob/export_history_effects/internal/transform/operation.go) (new functions)


cmd/export_effects.go
```diff
@@ -0,0 +1,84 @@
...
+		cmdLogger.SetLevel(logrus.InfoLevel)
+		endNum, strictExport, isTest, extra := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
+		cmdLogger.StrictExport = strictExport
+		startNum, path, limit := utils.MustArchiveFlags(cmd.Flags(), cmdLogger)
+		gcsBucket, gcpCredentials := utils.MustGcsFlags(cmd.Flags(), cmdLogger)
+		env := utils.GetEnvironmentDetails(isTest)
+
+		operations, err := input.GetOperations(startNum, endNum, limit, env)
+		if err != nil {
+			cmdLogger.Fatal("could not read effects: ", err)
```
Shaptic, line 26: operations, not effects?
```suggestion
			cmdLogger.Fatalf("could not read operations in [%d, %d] (limit=%d): %v", startNum, endNum, limit, err)
```

README.md
```diff
@@ -162,6 +145,34 @@ This command has two modes: bounded and unbounded.
...
+This command exports operations within the provided range.
+
+#### export_effects
+
+```bash
+> stellar-etl export_operations --start-ledger 1000 \
+--end-ledger 500000 --output exported_effects.txt
+```
+
+This command exports effects within the provided range.
```
Shaptic, line 173: Why did this block move? It seems like there's still an `--end-ledger`, meaning it's not an unbounded range like the new heading implies.

cmd/export_effects.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+
+		outFile := mustOutFile(path)
+		numFailures := 0
+		totalNumBytes := 0
+		for _, transformInput := range operations {
+			effects, err := transform.TransformEffect(transformInput.Operation, transformInput.OperationIndex, transformInput.Transaction, transformInput.LedgerSeqNum)
+			if err != nil {
+				txIndex := transformInput.Transaction.Index
+				cmdLogger.LogError(fmt.Errorf("could not transform operation %d in transaction %d in ledger %d: %v", transformInput.OperationIndex, txIndex, transformInput.LedgerSeqNum, err))
```
Shaptic, line 36: bit of a nitpick, but why not use `cmdLogger.Errorf`?

internal/transform/effects.go
```diff
@@ -0,0 +1,1289 @@
...
+}
+
+func (e *effectsWrapper) addCreatePassiveSellOfferEffect() error {
+	result := e.operation.OperationResult()
+	source := e.operation.SourceAccount()
+
+	var claims []xdr.ClaimAtom
+
+	// KNOWN ISSUE:  stellar-core creates results for CreatePassiveOffer operations
+	// with the wrong result arm set.
```
Shaptic, line 485: Can you cite an issue, code snippet, or slack thread for this? Would be nice to have a reference!

internal/transform/schema.go
```diff
@@ -293,3 +293,146 @@ type SponsorshipOutput struct {
 	Operation      xdr.Operation
 	OperationIndex uint32
 }
+
+// // EffectOutput is a representation of an operation that aligns with the BigQuery table history_effects
```
Shaptic, line 297: nit: extra comment line 
```suggestion
// EffectOutput is a representation of an operation that aligns with the BigQuery table history_effects
```

Shaptic (commented):
Thanks for this! It's a pretty massive PR (~1500 LoC of additions and another ~3k+ of testing) so my biggest concern is not getting coverage for the whole thing. It'd be really nice if you could summarize what types of behavior the test cases cover (i.e. so we don't have to read & understand all 3k lines lol) so we can get a sense of the coverage. 

As a separate, but related comment, is there a chance we can use fuzzing to generate transactions/operations and get lots of effects?

README.md
lucaszanotelli: These commands were in History Archive section before that's why I moved them. To make it more clear, I added a "Stellar Core bounded mode only" at [this commit](https://github.com/stellar/stellar-etl/pull/168/commits/f953baf995b631c495137e8d2381485ae1e45fff).

internal/transform/effects.go
lucaszanotelli: About this and the main concern you addressed in the comment, I forgot to mention that [`transform/effects.go`](https://github.com/stellar/stellar-etl/blob/export_history_effects/internal/transform/effects.go) is virtually [`effects_processor.go`](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/effects_processor.go) from Horizon.

I'll edit this PR to link files from here to those source files from Horizon.

lucaszanotelli: About adding fuzzing support, I think it's a great idea. The test file is massive and confuse.
The only problem is that we're depending on `history_effects` table to recreate the `enriched_operations` table. Maybe we should discuss about that in Slack.

bartekn:
If the files are identical (I haven't actually use `diff` to compare them) it should work however my concern is this is a file duplicated mostly 1-1 from Horizon. In case of any bugs/changes we need to fix it in two places. And I'm not really sure what's the best was to fix this. Maybe we should move Horizon processors outside `internal` directory so `stellar-etl` could import it?

lucaszanotelli:
I totally agree that we should keep the changes at one place. If we could move processors and also [history](https://github.com/stellar/go/tree/master/services/horizon/internal/db2/history) from `internal` I would move any work done here to Horizon.

Except for commit [30cf371](https://github.com/stellar/stellar-etl/commit/30cf3719bc068c4cfeb8e7400ee7bc41a85a4e41) and [229e214](https://github.com/stellar/stellar-etl/commit/229e2142647735b0ceb8b7ee28e2112d106c66b0) the rest is the duplicated from Horizon.

`diff` will show more output since [history](https://github.com/stellar/go/tree/master/services/horizon/internal/db2/history) was also copied from Horizon and placed inside [`transform/schema.go`](https://github.com/stellar/stellar-etl/commit/e6397c2a6b30aa71e31e171a67952a6fd66198a0).

Shaptic:
Thanks for explaining that this stuff came from Horizon; I was suspicious but didn't realize it was basically 1:1. I think moving processors outside of Horizon per @bartekn's suggestion could be valuable for stability, but it might be a big lift since I'm sure they depend on plenty of other `internal/` stuff (e.g. `history` as mentioned)... It almost sounds like stellar-etl should be a subset of Horizon's codebase.

Shaptic (approved):
We can worry about code duplication later once the Horizon team has more bandwidth.

PR #169, opened 2022-11-21
State table verification

lucaszanotelli:
### Overview

This implements Horizon's `verify` package to `stellar-etl` command `export_ledger_entry_changes`. It only verifies entries from checkpoint ledgers (every 64 ledgers, every ~5 minutes).

### Changes

* Add `Raw*` fields types `AccountOutput`, `OfferOutput`, `PoolOutput`, `ClaimableBalanceOutput` and `TrustlineOutput`
* Add `LedgerIsCheckpoint()` function to `utils` package
* Add `TransformedOutput` type to `internal/transform/schema.go` to assist the Horizon's `verify` package implementation
* Refactor `internal/utils/verify.go` to work with `stellar-etl`
* Update unit tests to add new fields

Closes [GH#279](https://github.com/stellar/hubble/issues/279), [GH#280](https://github.com/stellar/hubble/issues/280), [GH#281](https://github.com/stellar/hubble/issues/281)


cmd/export_ledger_entry_changes.go
```diff
@@ -91,77 +97,120 @@ be exported.`,
...
+								x := verifyOutputs[actualLedger]
+								x.Liquidity_pools = append(x.Liquidity_pools, pool)
+								verifyOutputs[actualLedger] = x
+							}
 						}
 					}
 				}
 
+				for checkpointLedgers := range verifyOutputs {
+					_, err := verify.VerifyState(ctx, verifyOutputs[checkpointLedgers], archive, checkpointLedgers, verifyBatchSize)
```
sydneynotthecity, line 208: Do you know where the `verifyBatchSize` comes from? Do we know that 50000 is big enough?

internal/transform/claimable_balance.go
```diff
@@ -47,11 +47,13 @@ func TransformClaimableBalance(ledgerChange ingest.Change) (ClaimableBalanceOutp
 
 	transformed := ClaimableBalanceOutput{
 		BalanceID:          balanceID,
+		Asset:              balanceEntry.Asset,
```
sydneynotthecity, line 50: Do these new columns get added to tables in Bigquery? What is the reasoning behind adding the elements here?

sydneynotthecity (approved):
Couple of questions regarding the newly added elements in the Output structs, otherwise looks good!

cmd/export_ledger_entry_changes.go
lucaszanotelli: `verifyBatchSize` value came from [here](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/verify.go#L23). About the second question, I don't know the answer. Maybe @bartekn could you help us with that?

internal/transform/claimable_balance.go
lucaszanotelli: These elements were added to overcome types incompatibility between functions from `verify` package and `stellar-etl` internal types. Its `json:` struct tags were given the special `"-"` value that makes the `encoding/json` package ignore parsing those fields. All new fields with the `Raw` prefix follows the same logic so it answers the question above.

internal/utils/verify/main.go
```diff
@@ -0,0 +1,227 @@
+// lucas zanotelli dos santos 2022-11-18
+// this package was copied from https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/verify/main.go
```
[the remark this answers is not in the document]
lucaszanotelli: This was a direct copy. There's no changes.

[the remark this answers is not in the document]
bartekn: I think we should move it out of `internal` package of Horizon to `github.com/stellar/go/ingest/verify`. Once there are new ledger entries only one place will have to be updated.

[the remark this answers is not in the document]
sydneynotthecity: +1 I agree that the `verify` package should be made public. I can write an issue in the Horizon repo for that and we can refactor `stellar-etl` once that package is public.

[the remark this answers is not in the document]
bartekn: https://github.com/stellar/go/pull/4709

internal/utils/verify/verify.go
```diff
@@ -0,0 +1,470 @@
...
+	if err != nil {
+		return false, errors.Wrap(err, "addClaimableBalanceToStateVerifier failed")
+	}
+
+	err = addLiquidityPoolsToStateVerifier(ctx, verifier, transformedOutput.Liquidity_pools)
+	if err != nil {
+		return false, errors.Wrap(err, "addLiquidityPoolsToStateVerifier failed")
+	}
+
+	return true, nil
```
bartekn, line 76: There should be `Verify()` call after `Write`'ing all ledger entries.

cmd/export_ledger_entry_changes.go
bartekn: This value can be changed. It's basically as big as possible but small enough to not cause OOM crash.

internal/utils/verify/main.go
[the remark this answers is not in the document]
bartekn: @lucaszanotelli the `verify` package is now public. You can remove the copy of the files and import `github.com/stellar/go/ingest/verify`.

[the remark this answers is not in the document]
sydneynotthecity: thanks @bartekn!

internal/utils/verify/verify.go
lucaszanotelli: @bartekn calling `Verify()` against `StateVerifier.currentEntries` makes sense when these entries come from the Captive Core? I'm struggling to implement this method as it is because the logic I used in this file is different than [the original one](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/verify.go).

If implementing this is really necessary, could you shed some light on how to?

sydneynotthecity (approved):
Looks good and ready for release

PR #170, opened 2022-11-21
Fix export_effects command

lucaszanotelli:
### Bug Overview

The `export_effects` command is duplicating records due to wrong code refactoring.

### Bug Fix

Get transactions instead of operations in `cmd/export_effects.go` and adapt `internal/transform/effects.go` code to create effects from that input.


sydneynotthecity (approved):
Looks good!

PR #172, opened 2022-12-09
Fix Dockerfile

lucaszanotelli:
I've updated the `stellar-etl` Docker image build to a two-stage build.

In the first stage it uses the `golang` container to build the `stellar-etl`. Then it calls `stellar-core` container to be the runtime environment and copies the needed files from build container to the runtime one.

This approach also contributes to an image size reduction. This builds an image about 80% lighter than the previous build.

Closes [stellar/hubble#320](https://github.com/stellar/hubble/issues/320)


sydneynotthecity (approved):
The fix works! Thanks @lucaszanotelli I was able to build the image locally and run some commands.

PR #174, opened 2022-12-13
Fix WORKDIR path

lucaszanotelli:
Remove the `/docker` append to keep the`WORKDIR` the same as the previous builds.


sydneynotthecity (approved):
LGTM

sydneynotthecity (approved):

PR #175, opened 2022-12-20
undo modifications

lucaszanotelli:
wanted to deploy history_transaction but it modifications to history_operations also.

basicaly undo pull request #164

PR #176, opened 2023-02-02
Revert "undo modifications"

sydneynotthecity:
Reverts stellar/stellar-etl#175


chowbao (approved):

internal/transform/schema.go
```diff
@@ -102,7 +102,7 @@ type OperationOutput struct {
 	SourceAccount      string                 `json:"source_account"`
 	SourceAccountMuxed string                 `json:"source_account_muxed,omitempty"`
 	Type               int32                  `json:"type"`
-	ApplicationOrder   int32                  `json:"application_order"`
+	TypeString         string                 `json:"type_string"`
```
tsachiherman, line 105: it's a bit wierd to see type and typestring within the same struct. Maybe in the future we could consider using type and subtype ?

PR #183, opened 2023-03-15
Add unique ID to assets

lucaszanotelli:
This PR features the use of the FarmHash hash function to generate an unique ID for assets.

The ID has type INT64 (8 bytes in BigQuery) and takes as input the asset code, issuer and type concatenated.

This new field was added to all tables that contains assets.


Shaptic (commented):
Couple of notes:

 - Is the goal here to easily and deterministically create relationships between rows based on the asset involved without string concatenation & comparison?
 - Any specific reason to use `FarmHash` over something built-in like sha256?
 - I'm not a huge fan of negative IDs - you could cast them to `uint64` to keep them positive? Just a thought

lucaszanotelli:
> * Is the goal here to easily and deterministically create relationships between rows based on the asset involved without string concatenation & comparison?

Yes, it is. To be more specific (I should've included this in the PR description) the goal is to reduce bytes processed by the query that generates the [`enriched_meaningful_history_operations`](https://github.com/stellar/stellar-etl-airflow/blob/master/dags/queries/enriched_meaningful_history_operations.sql) table.
This was already tested and validated. I've started a [thread in Slack](https://stellarfoundation.slack.com/archives/C03GQQ8GLBW/p1678300589898849) with the tests results, if you to know want more details of the methodology.

> * Any specific reason to use `FarmHash` over something built-in like sha256?

There is a couple of interconnected reasons. I wanted a hash function that's already build into BigQuery (so we can replicate it in SQL) and I wanted this column to be as small as possible — since the goal is to reduce bytes processed.
BQ has [`sha256`](https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions#sha256) build-in but it's output is 32 bytes.[`FarmHash`](https://cloud.google.com/bigquery/docs/reference/standard-sql/hash_functions#farm_fingerprint) has the smallest footprint: only 8 bytes.

> * I'm not a huge fan of negative IDs - you could cast them to `uint64` to keep them positive? Just a thought.

From a data size point of view, I don't think it's a good idea to switch to `uint64`.
The column size will double when doing this ([it will be type NUMERIC instead of INT64](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#data_type_sizes)), so we will spend more on table storage and scan more data when running the query.
Can you be more specific why you don't like the ID as `int64`?

internal/transform/asset.go
```diff
@@ -44,11 +45,14 @@ func transformSingleAsset(asset xdr.Asset) (AssetOutput, error) {
...
 	}
 
+	farmAssetID := FarmHashAsset(outputAssetCode, outputAssetIssuer, outputAssetType)
+
 	return AssetOutput{
 		AssetCode:   outputAssetCode,
 		AssetIssuer: outputAssetIssuer,
 		AssetType:   outputAssetType,
 		AssetID:     outputAssetID,
+		ID:          farmAssetID,
```
sydneynotthecity, line 55: Having two asset ids on the `history_assets` table might be confusing. We might want to consider dropping the original `AssetID` from the table in the future.

internal/transform/operation.go
```diff
@@ -279,6 +279,7 @@ func addAssetDetailsToOperationDetails(result map[string]interface{}, asset xdr.
 
 	result[prefix+"asset_code"] = code
 	result[prefix+"asset_issuer"] = issuer
+	result[prefix+"asset_id"] = FarmHashAsset(code, issuer, assetType)
```
sydneynotthecity, line 282: Is this the only `details` field that needs an asset id? What about trade operations or liquidity pool operations where the asset fields are `selling_asset_code` or `asset_a_code` etc?

Curious if this field is the biggest impact or if we need the other asset fields to have the asset id as well.

sydneynotthecity: Nevermind, I see this applies for all asset fields

internal/transform/operation_test.go
```diff
@@ -632,6 +633,7 @@ func makeOperationTestOutputs() (transformedOperations []OperationOutput) {
 				"selling_asset_code":   "USDT",
 				"selling_asset_type":   "credit_alphanum4",
 				"selling_asset_issuer": hardCodedDestAccountAddress,
+				"selling_asset_id":     int64(-8205667356306085451),
 				"buying_asset_type":    "native",
```
sydneynotthecity, line 637: Does `XLM` not get an id assigned? I would expect that even `XLM` should have an id.

sydneynotthecity (commented):
Two questions:

1. Should `XLM` also have an `asset_id`? Otherwise how are we including records where `XLM` was involved in an operation in our joins? I noticed that none of the test cases included an asset id if the asset type was `native`. 
2. How do you plan to update the existing tables to include the asset id?

Shaptic:
Thanks for the detailed answers!

>  Can you be more specific why you don't like the ID as int64?

Honestly just because negative IDs feel weird to me, hah. If it makes sense to keep them that way for technical reasons, I'm all for it :+1:

sydneynotthecity:
> Honestly just because negative IDs feel weird to me, hah. If it makes sense to keep them that way for technical reasons, I'm all for it

I tend to agree with your sentiment regarding negative ids. In this particular case, we're trying to reduce a large query cost in Bigquery that loads the `enriched_history_operations` table. Using this solution will reduce the total bytes scanned in the query by an estimated 97% 🤯

internal/transform/operation_test.go
lucaszanotelli: You're right. I'll add this right now.

internal/transform/asset.go
lucaszanotelli: I totally agree. Two IDs doesn't make any sense. We should drop it as soon as possible.

lucaszanotelli:
@sydneynotthecity since the XLM has no type and code, the id will always be the same value. So I hardcoded the value instead of calling the FarmHash function. Do you think is ok?

sydneynotthecity (approved):
LGTM Let's make sure to coordinate the release with `stellar-etl-airflow` changes so that we don't break the tables

PR #185, opened 2023-04-14
Update readme and docs

chowbao:
Update readme for stellar-etl

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ The Stellar-ETL is a data pipeline that allows users to extract data from the hi
       - [export_trades](#export_trades)
 	- [Stellar Core Commands](#stellar-core-commands)
 	  - [export_ledger_entry_changes](#export_ledger_entry_changes)
-	  - [export_orderbooks](#export_orderbooks)
-	- [Utility Commands](#utility-commands)
+      - [export_orderbooks (unsupported)](#export_orderbooks-unsupported)
+	  - [Utility Commands](#utility-commands)
 	  - [get_ledger_range_from_times](#get_ledger_range_from_times) 
 - [Schemas](#schemas)
 - [Extensions](#extensions)
@@ -66,7 +66,7 @@ The Stellar-ETL is a data pipeline that allows users to extract data from the hi
    - [export_trades](#export_trades)
  - [Stellar Core Commands](#stellar-core-commands)
    - [export_ledger_entry_changes](#export_ledger_entry_changes)
-   - [export_orderbooks](#export_orderbooks)
+   - [export_orderbooks (unsupported)](#export_orderbooks-unsupported)
  - [Utility Commands](#utility-commands)
    - [get_ledger_range_from_times](#get_ledger_range_from_times)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -102,12 +161,74 @@ These commands export information using the history archives. This allows users
...
+## **Stellar Core Commands**
 
 These commands require a Stellar Core instance that is v15.0.0 or later. The commands use the Core instance to retrieve information about changes from the ledger. These changes can be in the form of accounts, offers, or trustlines.
 
 As the Stellar network grows, the Stellar Core instance has to catch up on an increasingly large amount of information. This catch-up process can add some overhead to the commands in this category. In order to avoid this overhead, run prefer processing larger ranges instead of many small ones, or use unbounded mode.
-#### export_ledger_entry_changes
+
+<br>
+
+### **export_ledger_entry_changes**
```
sydneynotthecity, line 231: nit: update below to just say "flags can filter which ledger entry types are exported" and remove the "three data type flags" that way as we add new ledger entry types we don't need to keep updating

README.md
```diff
@@ -120,13 +241,15 @@ Changes are exported in batches of a size defined by the `batch-size` flag. By d
...
  If both a start and end ledger are provided, then the command runs in a bounded mode. This means that once all the ledgers in the range are processed and exported, the command shuts down.
  
-##### Unbounded
+#### **Unbounded**
 If only a start ledger is provided, then the command runs in an unbounded fashion starting from the provided ledger. In this mode, the Stellar Core connects to the Stellar network and processes new changes as they occur on the network. Since the changes are continually exported in batches, this process can be continually run in the background in order to avoid the overhead of closing and starting new Stellar Core instances.
 
-#### export_orderbooks
+<br>
+
+### **export_orderbooks**
```
sydneynotthecity, line 252: Drop all references to orderbooks. This was an experimental feature that we've never used.

sydneynotthecity: Or mention that it's experimental and unsupported

README.md
```diff
@@ -211,5 +313,4 @@ In general, in order to add new commands, you need to add these files:
 	- This file will contain the methods needed to transform the extracted data into a form that is suitable for BigQuery.
 	- The struct definition for the transformed object should be stored in `schemas.go` in the `internal/transform` folder.
 
-A good number of common methods are already written and stored in the `util` package.
-
+A good number of common methods are already written and stored in the `util` package.
```
sydneynotthecity, line 316: is it worth listing any of common methods?

sydneynotthecity (approved):
Overall looks really good. I have a couple of suggestions and nits.

README.md
chowbao: Marked it as unsupported

README.md
chowbao: I don't think so. There are a lot of functions and I think most of them are pretty self explanatory or have a comment/description

README.md
```diff
@@ -2,98 +2,157 @@
...
-#### export_accounts
+<br>
+
+### **export_accounts**
 
 ```bash
 > stellar-etl export_accounts --end-ledger 500000 --output exported_accounts.txt
 ```
 
 This command exports accounts, starting from the genesis ledger and ending at the ledger determined by `end-ledger`. This command exports the point-in-time state of accounts, meaning that the exported data represents the account information as it was at `end-ledger`.
```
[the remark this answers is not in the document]
chowbao: Hmm I updated the descriptions with the same ones in the .go files. But I think they might actually be point-in-time states. Or do I have the definition of all history wrong?

From what I can tell is all these export_* does the following:
* Start from genesis (ledger 0)
* Go to `--end-ledger`
* Only the latest data before or on `end-ledger` will be printed
** For example: export_account outputs only the latest entry up till end-ledger for any unique account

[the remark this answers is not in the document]
sydneynotthecity: Yes, you are right. we are using different ingest readers. This uses `NewCheckpointChangeReader` found [here](https://github.com/stellar/stellar-etl/blob/f5ce5bebe2813a2decb40c871f4955dd89af669a/internal/input/bucketlist_entries.go#L39)

PR #187, opened 2023-07-11
Added ledger_closed_at values to export results

chowbao:
Changes to support adding ledger_closed_at values to export results


cmd/export_ledger_entry_changes.go
```diff
@@ -99,7 +99,7 @@ be exported.`,
 								cmdLogger.LogError(fmt.Errorf("unable to identify changed accounts: %v", err))
 								continue
 							} else if changed {
-								acc, err := transform.TransformAccount(change)
+								acc, err := transform.TransformAccount(change, xdr.LedgerCloseMeta{})
```
sydneynotthecity, line 102: is this passing an empty `xdr.LedgerCloseMeta{}`?

internal/input/bucketlist_entries.go
```diff
@@ -60,3 +62,33 @@ func readBucketList(archive historyarchive.ArchiveInterface, checkpointSeq uint3
...
+		},
+	)
+
+	ledgerCloseMeta, err := backend.GetLedger(ctx, end)
+	if err != nil {
+		return xdr.LedgerCloseMeta{}, errors.Wrap(err, "error getting ledger from the backend")
+	}
+
+	return ledgerCloseMeta, nil
+}
```
sydneynotthecity, line 94: We don't need to address now, but I think it would be better for us to write a new interface in `utils/main.go` that will instantiate a new CaptiveCore and just use the GetLedger method for retrieving the `LedgerCloseMeta`. This current design is clunky and we do not have a consistent captive core interface.

internal/transform/account.go
```diff
@@ -21,6 +21,11 @@ func TransformAccount(ledgerChange ingest.Change) (AccountOutput, error) {
 		return AccountOutput{}, fmt.Errorf("Could not extract account data from ledger entry; actual type is %s", ledgerEntry.Data.Type)
 	}
 
+	outputCloseTime, err := utils.TimePointToUTCTimeStamp(ledgerCloseMeta.MustV0().LedgerHeader.Header.ScpValue.CloseTime)
```
sydneynotthecity, line 24: We need this to support both Meta types: `V0` and `V1`. Can you refactor to also return the `CloseTime` for `MustV1()`?

sydneynotthecity: I would also consider making this a method in `utils` so you can just pass LedgerCloseMeta from each transform to return the closedAt time.

internal/transform/asset.go
sydneynotthecity, line 1: You can drop the changes for the `history_assets` table. As discussed, we will not be adding the `closed_at` timestamp to that table since it doesn't make sense.

internal/transform/claimable_balance.go
```diff
@@ -36,7 +36,7 @@ func TransformClaimableBalance(ledgerChange ingest.Change) (ClaimableBalanceOutp
 		return ClaimableBalanceOutput{}, fmt.Errorf("Invalid balanceId in op: %d", uint32(ledgerEntry.LastModifiedLedgerSeq))
 	}
 	outputFlags := uint32(balanceEntry.Flags())
-	outputAsset, err := transformSingleAsset(balanceEntry.Asset)
+	outputAsset, err := transformSingleAsset(balanceEntry.Asset, LedgerCloseMeta)
```
sydneynotthecity, line 39: I don't think it makes sense to extract the ledger close timestamp in this method. The timestamp is tied to the ledger close, not asset. Is it possible to extract the timestamp separately? I wouldn't expect the `transformSingleAsset` to return the ledger closed timestamp without good reason.

sydneynotthecity (commented):
I have a couple of suggestions. Once they are addressed, will approve the PR. 

- since we do not need the `closed_at` timestamp in the `history_assets` table, I suggest removing from the `assets` code. Including removing from `transformSingleAsset` method. The code should be moved into its own method and should not be a part of that method.
- the current code only accounts for `LedgerCloseMeta` of type `V0`. I suggest writing a method that will handle both types, `V0` and `V1`. That will make it easier to pass `ledgerCloseMeta` from each processor anyway.

internal/utils/main.go
```diff
@@ -562,3 +562,48 @@ func GetEnvironmentDetails(isTest bool) (details EnvironmentDetails) {
...
+
+	return ledgerCloseMeta, nil
+}
+
+func GetCloseTimeV(ledger xdr.LedgerCloseMeta, v bool) (time.Time, error) {
+	if v == false {
+		return ExtractLedgerCloseTime(ledger.V0.LedgerHeader)
+	}
+	return ExtractLedgerCloseTime(ledger.V1.LedgerHeader)
+}
```
sydneynotthecity, line 609: I think we should wait until the latest xdr changes are merged into the master branch. There will be some new functions to help parse different `LedgerCloseMeta` types, like this `LedgerHeaderHistoryEntry()` [here](https://github.com/stellar/go/blob/soroban-xdr-next/xdr/ledger_close_meta.go#L5-L16)

```
func (l LedgerCloseMeta) LedgerHeaderHistoryEntry() LedgerHeaderHistoryEntry {
	switch l.V {
	case 0:
		return l.MustV0().LedgerHeader
	case 1:
		return l.MustV1().LedgerHeader
	case 2:
		return l.MustV2().LedgerHeader
	default:
		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
	}
}

```


Then you'll just be able to pass a generic `LedgerHeader` and extract the ledger close time regardless of ledger close type.

internal/utils/main.go
```diff
@@ -562,3 +562,48 @@ func GetEnvironmentDetails(isTest bool) (details EnvironmentDetails) {
 		return details
 	}
 }
+
+type CaptiveCore interface {
```
sydneynotthecity, line 566: Thanks for making this interface, much better than previous code

sydneynotthecity (commented):
Changes are looking good, thank you for refactoring the CaptiveCore backend so that we're only defining that once.

One additional feedback: 
We should only have one `closed_at` timestamp per transform. The `LedgerCloseMeta` entry can only be `V0` or `V1` or `V2`...etc. There will never be a scenario when a single `LedgerCloseMeta` will share multiple types. Therefore we should _switch_ between `LedgerHeaderHistoryEntry` based on the V type set in `LedgerCloseMeta.V`. Something like this: 

```
func getCloseTime(lcm LedgerCloseMeta) TimestampUTC int {
	switch l.V {
	case 0:
		return ExtractLedgerCloseTime(l.MustV0().LedgerHeader)
	case 1:
		return ExtractLedgerCloseTime(l.MustV1().LedgerHeader)
	case 2:
		return ExtractLedgerCloseTime(l.MustV2().LedgerHeader)
	default:
		panic(fmt.Sprintf("Unsupported LedgerCloseMeta.V: %d", l.V))
	}
}
```
Code is just an example, I haven't tested and don't know if it's correct.

internal/utils/main.go
sydneynotthecity: Update: we'll leave this as a to-do to refactor with the testnet release.

sydneynotthecity:
@cayod sorry for all of the back and forth on this PR. @chowbao and I would like this released _before_ our big Soroban release to reduce risk. Can you:

- revert the `stellar/go` library version to the prior version (without changes to `LedgerHeaderHistoryEntry`)
- revert the  `GetCloseTime()` method to your prior logic 

at that point the PR can be marked as Ready for review and we can work on releasing.

sydneynotthecity (approved):
LGTM

PR #188, opened 2023-07-14
Soroban support futurenet

chowbao:
diff soroban futurenet changes with master


internal/transform/config_setting.go
```diff
@@ -0,0 +1,164 @@
...
+	configSetting, ok := ledgerEntry.Data.GetConfigSetting()
+	if !ok {
+		return ConfigSettingOutput{}, fmt.Errorf("Could not extract config setting from ledger entry; actual type is %s", ledgerEntry.Data.Type)
+	}
+
+	configSettingId := configSetting.ConfigSettingId
+
+	contractMaxSizeBytes, ok := configSetting.GetContractMaxSizeBytes()
+
+	contractCompute, ok := configSetting.GetContractCompute()
```
sydneynotthecity, line 33: do we need to catch any errors if they occur? Would it be better to use a `MustContractMaxSizeBytes`?

sydneynotthecity (commented):
Reviewed a portion of the code, more to come

internal/transform/config_setting.go
chowbao: I believe the problem with the Must* functions is they `panic` which basically stops everything.
I removed the checks because most of the time most fields won't be set. Maybe I could refactor this to more like a switch statement with a struct of default null/zero values.

internal/transform/contract_code.go
```diff
@@ -0,0 +1,55 @@
...
+	contractCodeExpirationLedgerSeq := contractCode.ExpirationLedgerSeq
+
+	transformedPool := ContractCodeOutput{
+		ContractCodeHash:                contractCodeHash,
+		ContractCodeExtV:                int32(contractCodeExtV),
+		ContractCodeCode:                contractCodeCode,
+		ContractCodeExpirationLedgerSeq: uint32(contractCodeExpirationLedgerSeq),
+		LastModifiedLedger:              uint32(ledgerEntry.LastModifiedLedgerSeq),
+		LedgerEntryChange:               uint32(changeType),
+		Deleted:                         outputDeleted,
```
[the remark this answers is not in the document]
chowbao: I believe it has to be derived although I'm still not sure how to tell how evictions will work (different than expiration)

internal/transform/contract_data.go
```diff
@@ -0,0 +1,481 @@
...
+	metadataSym        = xdr.ScSymbol("METADATA")
+	metadataNameSym    = xdr.ScSymbol("name")
+	metadataSymbolSym  = xdr.ScSymbol("symbol")
+	adminSym           = xdr.ScSymbol("Admin")
+	issuerSym          = xdr.ScSymbol("issuer")
+	assetCodeSym       = xdr.ScSymbol("asset_code")
+	alphaNum4Sym       = xdr.ScSymbol("AlphaNum4")
+	alphaNum12Sym      = xdr.ScSymbol("AlphaNum12")
+	decimalSym         = xdr.ScSymbol("decimal")
+	assetInfoSym       = xdr.ScSymbol("AssetInfo")
```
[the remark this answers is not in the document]
chowbao: No idea. It was what was defined in Horizon and I'm assuming Horizon got it from core

internal/transform/config_setting.go
```diff
@@ -0,0 +1,164 @@
...
+		TempRentRateDenominator:         int64(tempRentRateDenominator),
+		MaxEntriesToExpire:              uint32(maxEntriesToExpire),
+		BucketListSizeWindowSampleSize:  uint32(bucketListSizeWindowSampleSize),
+		EvictionScanSize:                uint64(evictionScanSize),
+		LedgerMaxTxCount:                uint32(ledgerMaxTxCount),
+		BucketListSizeWindow:            bucketListSizeWindow,
+		LastModifiedLedger:              uint32(ledgerEntry.LastModifiedLedgerSeq),
+		LedgerEntryChange:               uint32(changeType),
+		Deleted:                         outputDeleted,
+	}
```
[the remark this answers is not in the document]
sydneynotthecity: This suggesting was harder to add than we originally thought. Have to either refactor the `ingest` library or our ingestion process to accommodate. Should not be included in this PR

internal/transform/contract_code.go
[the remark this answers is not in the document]
sydneynotthecity: Outdated--there are evicted entries now

internal/input/operations.go
```diff
@@ -56,13 +60,17 @@ func GetOperations(start, end uint32, limit int64, env utils.EnvironmentDetails)
...
-		// txReader, err := ingest.NewLedgerTransactionReader(ctx, backend, publicPassword, seq)
 		changeReader, err := ingest.NewLedgerChangeReader(ctx, backend, env.NetworkPassphrase, seq)
 		if err != nil {
 			return []OperationTransformInput{}, err
 		}
 		txReader := changeReader.LedgerTransactionReader
 
+		ledgerCloseMeta, err := backend.GetLedger(ctx, seq)
+		if err != nil {
+			return nil, errors.Wrap(err, "error getting ledger from the backend")
```
sydneynotthecity, line 71: would it be helpful to return the sequence number with the error so we know where the error occurred?

internal/utils/main.go
```diff
@@ -235,10 +270,13 @@ func AddExportTypeFlags(flags *pflag.FlagSet) {
 	flags.BoolP("export-offers", "f", false, "set in order to export offer changes")
 	flags.BoolP("export-pools", "p", false, "set in order to export liquidity pool changes")
 	flags.BoolP("export-balances", "l", false, "set in order to export claimable balance changes")
+	flags.BoolP("export-contract-code", "", false, "set in order to export claimable balance changes")
+	flags.BoolP("export-contract-data", "", false, "set in order to export claimable balance changes")
+	flags.BoolP("export-config-settings", "", false, "set in order to export claimable balance changes")
 }
```
sydneynotthecity, line 276: nit: description is a copy/paste error. Do we also need short flags for the new arguments? Or should all of them be removed?

internal/transform/config_setting.go
```diff
@@ -0,0 +1,161 @@
...
+	"strconv"
+
+	"github.com/stellar/go/ingest"
+	"github.com/stellar/go/xdr"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+// TransformConfigSetting converts an config setting ledger change entry into a form suitable for BigQuery
+func TransformConfigSetting(ledgerChange ingest.Change) (ConfigSettingOutput, error) {
+	ledgerEntry, changeType, outputDeleted, err := utils.ExtractEntryFromChange(ledgerChange)
```
sydneynotthecity, line 14: Do we also need to add a switch statement in `ExtractEntryFromChange` to handle evicted (expired?) entries? If that `LedgerEntryChangeType` is different from deleted, I want to make sure we're handling it.

internal/transform/config_setting.go
```diff
@@ -0,0 +1,161 @@
...
+
+	configSetting, ok := ledgerEntry.Data.GetConfigSetting()
+	if !ok {
+		return ConfigSettingOutput{}, fmt.Errorf("Could not extract config setting from ledger entry; actual type is %s", ledgerEntry.Data.Type)
+	}
+
+	// LedgerEntryChange must contain a config setting state change to be parsed, otherwise skip
+	if ledgerEntry.Data.Type != xdr.LedgerEntryTypeConfigSetting {
+		return ConfigSettingOutput{}, nil
+	}
```
sydneynotthecity, line 27: Is this redundant given the `GetConfigSetting()` method? curious if this will ever trigger

internal/transform/config_setting.go
```diff
@@ -0,0 +1,161 @@
...
+	contractExecutionLanes, _ := configSetting.GetContractExecutionLanes()
+	ledgerMaxTxCount := contractExecutionLanes.LedgerMaxTxCount
+
+	bucketList, _ := configSetting.GetBucketListSizeWindow()
+	bucketListSizeWindow := make([]uint64, 0, len(bucketList))
+	for _, sizeWindow := range bucketList {
+		bucketListSizeWindow = append(bucketListSizeWindow, uint64(sizeWindow))
+	}
+
+	transformedPool := ConfigSettingOutput{
```
sydneynotthecity, line 99: nit: `transformedPool` is copy/pasta

internal/transform/config_setting.go
```diff
@@ -0,0 +1,161 @@
...
+		PersistentRentRateDenominator:   int64(persistentRentRateDenominator),
+		TempRentRateDenominator:         int64(tempRentRateDenominator),
+		MaxEntriesToExpire:              uint32(maxEntriesToExpire),
+		BucketListSizeWindowSampleSize:  uint32(bucketListSizeWindowSampleSize),
+		EvictionScanSize:                uint64(evictionScanSize),
+		LedgerMaxTxCount:                uint32(ledgerMaxTxCount),
+		BucketListSizeWindow:            bucketListSizeWindow,
+		LastModifiedLedger:              uint32(ledgerEntry.LastModifiedLedgerSeq),
+		LedgerEntryChange:               uint32(changeType),
+		Deleted:                         outputDeleted,
```
sydneynotthecity, line 145: Is it worthwhile to add a boolean for `Expired`? Or do you think the `LedgerEntryChange` is enough?

internal/transform/contract_code.go
```diff
@@ -0,0 +1,58 @@
...
+	// NOTE: Most likely don't need the binary code to be saved in BQ
+	//binaryCode, err := contractCode.Body.MarshalBinary()
+	//if err != nil {
+	//	return ContractCodeOutput{}, fmt.Errorf("Could not extract Val from contractData")
+	//}
+	//contractCodeCode := base64.StdEncoding.EncodeToString(binaryCode)
+
+	contractCodeExpirationLedgerSeq := contractCode.ExpirationLedgerSeq
+
+	transformedPool := ContractCodeOutput{
```
sydneynotthecity, line 47: nit: `transformedPool` is copy/pasta

internal/transform/contract_code.go
```diff
@@ -0,0 +1,58 @@
...
+		ContractCodeHash:                contractCodeHash,
+		ContractCodeExtV:                int32(contractCodeExtV),
+		ContractCodeExpirationLedgerSeq: uint32(contractCodeExpirationLedgerSeq),
+		ContractCodeEntryBodyType:       contractCodeEntryBodyType,
```
sydneynotthecity, line 51: Do you know if this field will ever be a domain value beyond `ContractEntryBodyTypeDataEntry`? Curious if this field is actually useful or if we're parsing the wrong part of the XDR. I queried `contract_code` in BQ and there was only a single value in this column.

internal/transform/contract_data.go
```diff
@@ -0,0 +1,503 @@
...
+	transformedPool := ContractDataOutput{
+		ContractId:                  contractDataContractId.HexString(),
+		ContractKeyType:             contractDataKeyType,
+		ContractDurability:          contractDataDurability,
+		ContractDataFlags:           uint32(contractDataDataFlags),
+		ContractExpirationLedgerSeq: uint32(contractDataExpirationLedgerSeq),
+		ContractDataAssetCode:       contractDataAssetCode,
+		ContractDataAssetIssuer:     contractDataAssetIssuer,
+		ContractDataBalanceHolder:   base64.StdEncoding.EncodeToString(contractDataBalanceHolder[:]),
+		ContractDataBalance:         contractDataBalance.String(),
```
sydneynotthecity, line 127: do we want to store the balance as a string?

internal/transform/contract_data.go
```diff
@@ -0,0 +1,503 @@
...
+	// Useful contract data values are flattened into their own columns
+	//valBinary, err := contractDataData.Val.MarshalBinary()
+	//if err != nil {
+	//	return ContractDataOutput{}, fmt.Errorf("Could not extract Val from contractData")
+	//}
+	//contractDataDataVal := base64.StdEncoding.EncodeToString(valBinary)
+
+	contractDataExpirationLedgerSeq := contractData.ExpirationLedgerSeq
+
+	transformedPool := ContractDataOutput{
```
sydneynotthecity, line 118: nit: `transformedPool`

internal/transform/contract_data.go
```diff
@@ -0,0 +1,503 @@
...
+	}
+
+	contractDataAssetCode, contractDataAssetIssuer := t.AssetFromContractData(ledgerEntry, passphrase)
+
+	contractDataBalanceHolder, contractDataBalance, _ := t.ContractBalanceFromContractData(ledgerEntry, passphrase)
+
+	if contractData.Key.Type.String() == "ScValTypeScvLedgerKeyNonce" {
+		// Is a nonce and should be discarded
+		return ContractDataOutput{}, nil
+	}
```
sydneynotthecity, line 86: should this be moved up in the function so that we're continuing out of none key types faster?

internal/transform/contract_data.go
```diff
@@ -0,0 +1,503 @@
...
+	if contractData.Key.Type != xdr.ScValTypeScvLedgerKeyContractInstance ||
+		contractData.Body.BodyType != xdr.ContractEntryBodyTypeDataEntry {
+		return "", ""
+	}
+	contractInstanceData, ok := contractData.Body.Data.Val.GetInstance()
+	if !ok || contractInstanceData.Storage == nil {
+		return "", ""
+	}
+
+	// we don't support asset stats for lumens
```
sydneynotthecity, line 180: What are the implications of Hubble not storing asset stats for lumens? Is that ok?

internal/transform/effects.go
```diff
@@ -1278,3 +1295,212 @@ func (e *effectsWrapper) addLiquidityPoolWithdrawEffect() error {
 	e.addMuxed(e.operation.SourceAccount(), EffectLiquidityPoolWithdrew, details)
 	return nil
 }
+
+// addInvokeHostFunctionEffects iterates through the events and generates
+// account_credited and account_debited effects when it sees events related to
+// the Stellar Asset Contract corresponding to those effects.
+func (e *effectsWrapper) addInvokeHostFunctionEffects(events []contractevents.Event) error {
+	if e.operation.network == "" {
+		return errors.New("invokeHostFunction effects cannot be determined unless network passphrase is set")
```
sydneynotthecity, line 1304: FYI the network passphrase needs to be passed to the transformEffects function. I don't think it's passed rn so this will always error

internal/transform/effects_test.go
sydneynotthecity, line 1: If you get time, adding a unit test for some of the contract events would be excellent

sydneynotthecity (commented):
Reviewed through the ledger transformation code, will review operations in a later setting

internal/transform/schema.go
```diff
@@ -445,3 +460,83 @@ type TestTransaction struct {
...
+	ContractDataBalanceHolder   string `json:"balance_holder"`
+	ContractDataBalance         string `json:"balance"` // balance is a string because it is go type big.Int
+	LastModifiedLedger          uint32 `json:"last_modified_ledger"`
+	LedgerEntryChange           uint32 `json:"ledger_entry_change"`
+	Deleted                     bool   `json:"deleted"`
+	//ContractKey                 string `json:"contract_key"`
+	//ContractDataVal             string `json:"contract_data_val"` // Useful contract data is flattened into own columns
+}
+
+// ContractCodeOutput is a representation of contract data that aligns with the Bigquery table soroban_contract_data
```
sydneynotthecity, line 482: nit: comment for `ContractCodeOutput` and `ConfigSettingOutput` need updating

cmd/export_ledger_entry_changes.go
```diff
@@ -44,8 +44,8 @@ be exported.`,
 		}
 
 		// If none of the export flags are set, then we assume that everything should be exported
-		if !exportAccounts && !exportOffers && !exportTrustlines && !exportPools && !exportBalances {
-			exportAccounts, exportOffers, exportTrustlines, exportPools, exportBalances = true, true, true, true, true
+		if !exportAccounts && !exportOffers && !exportTrustlines && !exportPools && !exportBalances && !exportContractCode && !exportContractData && !exportConfigSettings {
+			exportAccounts, exportOffers, exportTrustlines, exportPools, exportBalances, exportContractCode, exportContractData, exportConfigSettings = true, true, true, true, true, true, true, true
```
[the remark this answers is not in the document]
chowbao: These flags don't even do anything 🤣 
I can make them actually functional

internal/transform/config_setting.go
chowbao: Currently there is no evicted/expired state (at least from looking at the XDR). When it's added we can update this to handle it

cmd/export_ledger_entry_changes.go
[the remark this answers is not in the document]
sydneynotthecity: HA. Well in that case, are they even needed? Do you think they would be useful in the CLI?

internal/transform/config_setting.go
chowbao: What would the boolean for `Expired` be for? config_settings don't expire afaik

internal/transform/contract_code.go
chowbao: Unsure if it'll ever have anything different in the future. Removed for now

internal/transform/contract_data.go
chowbao: Balance needs to be a string I think because it is the `big.Int` type in golang. We would need to convert it to `BIGNUMERIC` in BQ when we ingest it

internal/transform/operation.go
```diff
@@ -917,6 +931,88 @@ func extractOperationDetails(operation xdr.Operation, transaction ingest.LedgerT
...
+		}
+	case xdr.OperationTypeBumpFootprintExpiration:
+		op := operation.Body.MustBumpFootprintExpirationOp()
+		details["type"] = "bump_footprint_expiration"
+		details["ledgers_to_expire"] = op.LedgersToExpire
+
+		transactionEnvelope := transaction.Envelope.MustV1()
+		details["contract_id"] = contractIdFromTxEnvelope(transactionEnvelope)
+		details["contract_code_hash"] = contractCodeHashFromTxEnvelope(transactionEnvelope)
+	case xdr.OperationTypeRestoreFootprint:
```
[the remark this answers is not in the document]
chowbao: Hmm it doesn't from looking at the XDR. I'm not sure what the operation flow would be but the expiration time would be set from `BumpFootprintExpirationOp`
So not sure if you need to call both operations to get it to work

internal/transform/operation.go
```diff
@@ -917,6 +931,88 @@ func extractOperationDetails(operation xdr.Operation, transaction ingest.LedgerT
...
+			details["type"] = "invoke_contract"
+
+			transactionEnvelope := transaction.Envelope.MustV1()
+			details["contract_id"] = contractIdFromTxEnvelope(transactionEnvelope)
+			details["contract_code_hash"] = contractCodeHashFromTxEnvelope(transactionEnvelope)
+
+			// Host function parameters -- not useful for data analysis
+			//args := op.HostFunction.MustInvokeContract()
+			//params := make([]map[string]string, 0, len(args))
+			//
```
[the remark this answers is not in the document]
chowbao: Removed params from both

internal/transform/contract_data.go
chowbao: Good catch. We do want to capture the native asset. Adjusted the check to get the asset and the balance from contract data

internal/transform/operation.go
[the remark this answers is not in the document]
chowbao: From [here](https://soroban.stellar.org/docs/fundamentals-and-concepts/state-expiration#examples)
Operation OperationTypeRestoreFootprint will add a buffer amount for the expiration ledger (set by config_settings) which will allow the user to use it or to call BumpFootprintExpirationOp on it to further increase the expiration ledger

cmd/export_ledger_entry_changes.go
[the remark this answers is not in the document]
sydneynotthecity: discussed that the flags are useful so we will keep them in the code for now

internal/transform/operation_test.go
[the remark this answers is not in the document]
chowbao: I tried adding in unit tests but it was a little more involved than it's worth at the moment. Can add in the future

sydneynotthecity (approved):
Ready to be merged for testnet release

jensenity:
Will there be a README for the soroban support released? Do I still need a run captive-core if i run in on futurenet?

internal/transform/operation.go
```diff
@@ -1600,11 +1601,13 @@ func contractIdFromContractData(ledgerKey xdr.LedgerKey) string {
 		return ""
 	}
 	contractIdHash, ok := contractData.Contract.GetContractId()
-	if ok {
-		return contractIdHash.HexString()
+	if !ok {
+		return ""
```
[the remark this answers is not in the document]
chowbao: > are there circumstances when we won't pull the contract id correctly for a contract invocation? 
There shouldn't be. Yes it's just to keep ingestion going in case something is wrong

chowbao:
> Will there be a README for the soroban support released? Do I still need a run captive-core if i run in on futurenet?

I can update the README. You will need to run captive-core for certain export_* (e.g. export_operations) but that should be automatic when running the `stellar-etl` command. Soroban is now also released on testnet and these changes will also be compatible with testnet.

PR #191, opened 2023-07-18
Rewrite historyarchive backend for export ledgers

sydneynotthecity:
The way that the `export_ledgers` job ingests ledgers is incorrect and introduces the job to operational risk.

Because ledger data can be parsed without a captive core, the original `ledgerbackend` read ledgers from history archive, but spoofed the existence of `LedgerCloseMeta` so that the code could reuse the same backend used in the other ETL jobs. This is risky because the `LedgerCloseMeta` is hardcoded and assumes that the data read followed a `V0` format for `LedgerCloseMeta`

This PR refactors the code to use a custom create backend, `historyarchive` and ingests the `Ledger` directly from the archive instead of passing through a spoofed `ledgerbackend`

### DIFF
--- a/internal/input/ledgers.go
+++ b/internal/input/ledgers.go
@@ -5,30 +5,30 @@ import (
 
 	"github.com/stellar/stellar-etl/internal/utils"
 
-	"github.com/stellar/go/xdr"
+	"github.com/stellar/go/historyarchive"
 )
 
 // GetLedgers returns a slice of ledger close metas for the ledgers in the provided range (inclusive on both ends)
-func GetLedgers(start, end uint32, limit int64, isTest bool, isFuturenet bool) ([]xdr.LedgerCloseMeta, error) {
+func GetLedgers(start, end uint32, limit int64, isTest bool, isFuturenet bool) ([]historyarchive.Ledger, error) {
 	env := utils.GetEnvironmentDetails(isTest, isFuturenet)
 	backend, err := utils.CreateBackend(start, end, env.ArchiveURLs)
 	if err != nil {
-		return []xdr.LedgerCloseMeta{}, err
+		return []historyarchive.Ledger{}, err
 	}
 
-	metaSlice := []xdr.LedgerCloseMeta{}
-	ctx := context.Background()
+	ledgerSlice := []historyarchive.Ledger{}
 	for seq := start; seq <= end; seq++ {
-		ledger, err := backend.GetLedger(ctx, seq)
+		ctx := context.Background()
+		ledger, err := backend.GetLedgerArchive(ctx, seq)
 		if err != nil {
-			return []xdr.LedgerCloseMeta{}, err
+			return []historyarchive.Ledger{}, err
 		}
 
-		metaSlice = append(metaSlice, ledger)
-		if int64(len(metaSlice)) >= limit && limit >= 0 {
+		ledgerSlice = append(ledgerSlice, ledger)
+		if int64(len(ledgerSlice)) >= limit && limit >= 0 {
 			break
 		}
 	}
 
-	return metaSlice, nil
+	return ledgerSlice, nil
 }
--- a/internal/transform/ledger.go
+++ b/internal/transform/ledger.go
@@ -7,57 +7,57 @@ import (
 	"github.com/stellar/stellar-etl/internal/toid"
 	"github.com/stellar/stellar-etl/internal/utils"
 
+	"github.com/stellar/go/historyarchive"
 	"github.com/stellar/go/xdr"
 )
 
-//TransformLedger converts a ledger from the history archive ingestion system into a form suitable for BigQuery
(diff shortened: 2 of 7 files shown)


internal/input/ledgers.go
```diff
@@ -5,30 +5,30 @@ import (
...
-		return []xdr.LedgerCloseMeta{}, err
+		return []historyarchive.Ledger{}, err
 	}
 
-	metaSlice := []xdr.LedgerCloseMeta{}
-	ctx := context.Background()
+	ledgerSlice := []historyarchive.Ledger{}
 	for seq := start; seq <= end; seq++ {
-		ledger, err := backend.GetLedger(ctx, seq)
+		ctx := context.Background()
```
chowbao, line 21: nit: does ctx need to be inside the for loop?

internal/transform/ledger.go
```diff
@@ -7,57 +7,57 @@ import (
...
-func TransformLedger(inputLedgerMeta xdr.LedgerCloseMeta) (LedgerOutput, error) {
+// TransformLedger converts a ledger from the history archive ingestion system into a form suitable for BigQuery
+func TransformLedger(inputLedger historyarchive.Ledger) (LedgerOutput, error) {
 
-	ledgerHeaderHistory := inputLedgerMeta.LedgerHeaderHistoryEntry()
-	ledgerHeader := ledgerHeaderHistory.Header
+	ledgerHeader := inputLedger.Header.Header
 
-	outputSequence := uint32(ledgerHeader.LedgerSeq)
+	outputSequence := uint32(inputLedger.Header.Header.LedgerSeq)
```
chowbao, line 19: nit: `inputLedger.Header.Header.LedgerSeq` --> `ledgerHeader.LedgerSeq`

internal/transform/ledger.go
```diff
@@ -7,57 +7,57 @@ import (
...
 
 	outputLedgerID := toid.New(int32(outputSequence), 0, 0).ToInt64()
 
-	outputLedgerHash := utils.HashToHexString(ledgerHeaderHistory.Hash)
-	outputPreviousHash := utils.HashToHexString(ledgerHeader.PreviousLedgerHash)
+	outputLedgerHash := utils.HashToHexString(inputLedger.Header.Hash)
+	outputPreviousHash := utils.HashToHexString(inputLedger.Header.Header.PreviousLedgerHash)
```
chowbao, line 24: nit: `inputLedger.Header.Header.PreviousLedgerHash` --> `ledgerHeader.PreviousLedgerHash`

internal/transform/ledger.go
```diff
@@ -113,18 +114,39 @@ func extractCounts(lcm xdr.LedgerCloseMeta) (transactionCount int32, operationCo
...
+		panic(fmt.Sprintf("Unsupported TransactionHistoryEntry.Ext: %d", transactionEntry.Transaction.Ext.V))
 	}
 }
 
+func getTransactionPhase(transactionPhase []xdr.TransactionPhase) (transactionEnvelope []xdr.TransactionEnvelope) {
+	transactionSlice := []xdr.TransactionEnvelope{}
+	for _, phase := range transactionPhase {
+		switch phase.V {
+		case 0:
+			for _, component := range *phase.V0Components {
```
chowbao, line 136: nit: Could use the xdr Must/GetV0Components() functions as an extra form of checking if xdr elements are correct or not

internal/transform/ledger_test.go
```diff
@@ -78,24 +87,24 @@ func makeLedgerTestOutput() (output LedgerOutput, err error) {
 		OperationCount:             10,
 		SuccessfulTransactionCount: 1,
 		FailedTransactionCount:     1,
-		TxSetOperationCount:        "13",
+		TxSetOperationCount:        "2",
```
chowbao, line 90: How was the original test 13? The hardCodedTxSet was always 2?

internal/utils/main.go
```diff
@@ -112,6 +112,37 @@ func CreateSampleResultMeta(successful bool, subOperationCount int) xdr.Transact
 	}
 }
 
+func CreateSampleResultPair(successful bool, subOperationCount int) xdr.TransactionResultPair {
```
chowbao, line 115: This is only used for testing right?

chowbao (approved):

internal/utils/main.go
sydneynotthecity: correct, we have to build a sample `TransactionResultPair` to parse in the unit tests

internal/input/ledgers.go
sydneynotthecity: moving it out

internal/transform/ledger_test.go
sydneynotthecity: hmmmm it should actually be 13. looking into this

PR #194, opened 2023-07-24
Fix failures with soroban effects/ops

sydneynotthecity:


sydneynotthecity:
@chowbao when you have a second, I think this needs to be merged unless you already fixed these failures

PR #198, opened 2023-09-27
create export-tx

cayod:
This PR creates an export_tx for a test of the creation of a full history_transactional table from tx data in BQ


sydneynotthecity (approved):
@cayod looks good, go ahead and merge when ready

PR #199, opened 2023-10-03
Fix diagnostic event parsing bug

chowbao:
Fixed diagnostic event parsing to account for events without a contract_id

### DIFF
--- a/internal/transform/diagnostic_events.go
+++ b/internal/transform/diagnostic_events.go
@@ -34,11 +34,11 @@ func TransformDiagnosticEvent(transaction ingest.LedgerTransaction, lhe xdr.Ledg
 	var transformedDiagnosticEvents []DiagnosticEventOutput
 
 	for _, diagnoticEvent := range transactionMeta.SorobanMeta.DiagnosticEvents {
+		var outputContractId string
+
 		outputInSuccessfulContractCall := diagnoticEvent.InSuccessfulContractCall
 		event := diagnoticEvent.Event
 		outputExtV := event.Ext.V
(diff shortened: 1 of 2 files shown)


internal/transform/diagnostic_events.go
```diff
@@ -51,6 +51,12 @@ func TransformDiagnosticEvent(transaction ingest.LedgerTransaction, lhe xdr.Ledg
 			continue
 		}
 
+		if event.ContractId != nil {
```
sydneynotthecity, line 54: Curious, in what scenarios would the `ContractId` be null? Would this be null during the original WASM upload to host?

sydneynotthecity (approved):
LGTM

internal/transform/diagnostic_events.go
chowbao: Yes that's one case but I also think it's unoptimized errors. 

Example transaction_hash: 559f7422e90d054d77cbc32a756af55ff9838ca077f75b2e8b8329971b9eb57c
The transaction failed because `invokeHostFunctionResourceLimitExceeded` but it doesn't give you the contract_id that was used because the resource error is associated with the transaction and not necessarily a problem with the contract being invoked.

Could it have given a contract_id? Yes but it doesn't 🤷‍♀️ 
Can we find out the contract_id? Probably yes, with joins to transaction_id --> history_operations --> contract_id

sydneynotthecity: Ah, that makes sense. I don't think we need to always find the contract_id, we can leave that up to the user if they need to trace

PR #200, opened 2023-10-11
fixed history_trade and liquidity_pool asset_id field

cayod:
This PR fixes the `asset_id` field in `history_trade` and `liquidity_pool` table.


internal/transform/liquidity_pool.go
```diff
@@ -36,15 +36,15 @@ func TransformPool(ledgerChange ingest.Change) (PoolOutput, error) {
 	}
 
 	var assetAType, assetACode, assetAIssuer string
-	err = cp.Params.AssetA.Extract(&assetAType, &assetACode, &assetAIssuer)
+	err = cp.Params.AssetA.Extract(&assetACode, &assetAIssuer, &assetAType)
```
sydneynotthecity, line 39: `Extract` actually takes this form: 
```
 Extract(typ interface{}, code interface{}, issuer interface{})
```

See [source code](https://pkg.go.dev/github.com/stellar/go/xdr#Asset.Extract)

sydneynotthecity (commented):
The `Extract()` method for `Asset` follows a different variable ordering than `FarmHashAsset`. I think some of the tests are failing due to this change (plus the IDs in the unit test need to be changed.)

I also noticed in the `asset.go` code that we create an additional hash using the method `hashAsset`.  We return the following output: 
```
return AssetOutput{
		AssetCode:   outputAssetCode,
		AssetIssuer: outputAssetIssuer,
		AssetType:   outputAssetType,
		AssetID:     outputAssetID,
		ID:          farmAssetID,
	}, nil
```

I think it would make more sense to switch the `ID` and `AssetID` to be consistent with naming on other tables: 
```
AssetID:     farmAssetID,
ID:               outputAssetID,
```

Once the unit tests are passing, I will approve.

sydneynotthecity (approved):
Looks good, thanks for the updates

PR #201, opened 2023-10-13
fixed mismatch type in assetOutput schema

cayod:
This PR fixes the mismatch in `assetOutput` schema.

To do this, we had to revert the previous changes that invert `asset_id` and `id` in `assetOutput`


sydneynotthecity (approved):
LGTM

PR #203, opened 2023-10-17
Revert usedb flag which is slow

sydneynotthecity:


chowbao (approved):

internal/input/changes.go
```diff
@@ -32,7 +32,7 @@ func PrepareCaptiveCore(execPath string, tomlPath string, start, end uint32, env
 			NetworkPassphrase:  env.NetworkPassphrase,
 			HistoryArchiveURLs: env.ArchiveURLs,
 			Strict:             true,
-			UseDB:              true,
+			UseDB:              false,
```
sreuland, line 35: was this still slow when BucketListDB was enabled, the setting is only conditionally auto-enabled in horizon 2.24 or greater and when the stellar core [binary reports at least 19.6 version stamp](https://github.com/stellar/go/blob/master/services/horizon/CHANGELOG.md#2240)

is the stellar core binary installed from a released debian package, to verify, if you run `stellar-core version` what does it report? the core binary only prints the semver version stamp when built from deb pkg, if it's built from source, then it won't return the expected semver version string and that would prevent the auto-enablement of bucketlistdb usage on captive core in horizon.

to have it be more deterministic, it can be forced on horizon 2.24 and up to always enable bucketlist db on captive core by adding it in the captive core toml conf file template you pass to horizon:

```
EXPERIMENTAL_BUCKETLIST_DB = true
EXPERIMENTAL_BUCKETLIST_DB_INDEX_PAGE_SIZE_EXPONENT = 12
```

just wanted to mention the corner case as it could quietly lead to disabled bucketlistdb, which leads to much more i/o on the disk based db.

PR #204, opened 2023-10-18
Remove ledger ops limit

chowbao:
MaxTxSetSize does not account for soroban operations. Instead of doing an explicit check for operation types vs MaxTxSetSize just remove the check from stellar-etl and add a data quality check downstream (also core should already be checking for this error)


sydneynotthecity (approved):
Thanks for removing. In the future, some flavor of this data quality check might be useful in dbt but no action required today

PR #208, opened 2023-11-02
Add closed_at to ledger entry changes tables

chowbao:
Add `closed_at` to `export_ledger_entry_changes` tables.
The standalone "from genesis" export_* use an empty `closed_at` because the export_* are unused and the `CheckpointChangeReader` doesn't offer an easy way to get the ledger_sequence or closeMeta

### DIFF
--- a/cmd/export_claimable_balances.go
+++ b/cmd/export_claimable_balances.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -37,7 +38,8 @@ var claimableBalancesCmd = &cobra.Command{
 		numFailures := 0
 		totalNumBytes := 0
 		for _, balance := range balances {
-			transformed, err := transform.TransformClaimableBalance(balance)
+			var closedAt time.Time
+			transformed, err := transform.TransformClaimableBalance(balance, closedAt)
 			if err != nil {
 				cmdLogger.LogError(fmt.Errorf("could not transform balance %+v: %v", balance, err))
 				numFailures += 1
--- a/cmd/export_claimable_balances.go
+++ b/cmd/export_claimable_balances.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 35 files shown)


cmd/export_claimable_balances.go
```diff
@@ -37,7 +37,8 @@ var claimableBalancesCmd = &cobra.Command{
 		numFailures := 0
 		totalNumBytes := 0
 		for _, balance := range balances {
-			transformed, err := transform.TransformClaimableBalance(balance)
+			var header xdr.LedgerHeaderHistoryEntry
```
sydneynotthecity, line 40: nit: should the variable definition move outside of the `for` loop?

internal/input/changes.go
```diff
@@ -107,19 +112,13 @@ func extractBatch(
...
 			}
-			// TODO: Add in ledger_closed_at; Update changeCompactors to also save ledger close time.
-			//   AddChange is from the go monorepo so it might be easier to just add a addledgerclose func after it
-			//txReader := changeReader.LedgerTransactionReader
-
-			//closeTime, err := utils.TimePointToUTCTimeStamp(txReader.GetHeader().Header.ScpValue.CloseTime)
-			//if err != nil {
-			//	logger.Fatal(fmt.Sprintf("unable to read close time for ledger %d: ", seq), err)
-			//}
+			header = changeReader.LedgerTransactionReader.GetHeader()
```
sydneynotthecity, line 121: nice find :) was this method easy to find? Or would it be helpful to add to the `ingest` documentation for ease of use in the future?

sydneynotthecity (approved):
Small nit on where `header` is defined in some of the `cmd/` files but everything is good as is

internal/input/changes.go
chowbao: It wasn't necessarily obvious because of all the different reader nesting. But when you know it exists it is easy. Probably would be useful in the ingest documentation. It would be nice to get a reader/backend to xdr.* mapping like `{Reader: [xdr.TxMeta, xdr.CloseMeta, xdr.LedgerHeader, etc...]}` without having to look through the xdr

sydneynotthecity: That's a good idea. Let's discuss with the rest of platform whether it makes sense to change the backend xdr mapping to be clearer.

PR #209, opened 2023-12-07
Updates for 20.0.0 release

chowbao:
Updates for futurenet release of 20.0.0

`Expiration` is renamed to `ttl`


docker/Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY . .
 RUN go build -v -o /usr/local/bin ./...
 
 # stage 2: runtime enviroment
-FROM stellar/stellar-core:20.0.0-1504.rc2.22088c1f2.focal
+FROM chowbao/stellar-core:20.0.0-1615.617729910.focal
```
sydneynotthecity, line 16: ah, still no public docker image?

internal/transform/ledger.go
```diff
@@ -83,8 +83,6 @@ func TransactionProcessing(l xdr.LedgerCloseMeta) []xdr.TransactionResultMeta {
 		return l.MustV0().TxProcessing
 	case 1:
 		return l.MustV1().TxProcessing
-	case 2:
-		return l.MustV2().TxProcessing
```
sydneynotthecity, line 87: did transactionmetav2 get deleted? Confirming this was intentional.

internal/transform/operation.go
```diff
@@ -1003,10 +1003,10 @@ func extractOperationDetails(operation xdr.Operation, transaction ingest.LedgerT
...
 			panic(fmt.Errorf("unknown host function type: %s", op.HostFunction.Type))
 		}
-	case xdr.OperationTypeBumpFootprintExpiration:
-		op := operation.Body.MustBumpFootprintExpirationOp()
-		details["type"] = "bump_footprint_expiration"
-		details["ledgers_to_expire"] = op.LedgersToExpire
+	case xdr.OperationTypeExtendFootprintTtl:
+		op := operation.Body.MustExtendFootprintTtlOp()
+		details["type"] = "extend_footprint_ttl"
+		details["extend_to"] = op.ExtendTo
```
sydneynotthecity, line 1009: Since the keys in the object were renamed, how do you think making the updates to the `details` object in BQ is going to go? Curious if we'll be able to drop the old keys, or if we will just have to append the new fields

internal/transform/ttl.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+
+	"github.com/stellar/go/ingest"
+	"github.com/stellar/go/strkey"
+	"github.com/stellar/go/xdr"
+	"github.com/stellar/stellar-etl/internal/utils"
+)
+
+// TransformConfigSetting converts an config setting ledger change entry into a form suitable for BigQuery
```
sydneynotthecity, line 12: nit: copy/pasta error

sydneynotthecity (approved):
Changes look good. Feel free to merge when ready. 

While the raw data structure is named `ttl`, we might want to consider naming marts tables something a little more descriptive than ttl. 🤷‍♀️ that or we'll need to have robust docs so that end users know exactly what that means

internal/transform/operation.go
chowbao: It just appends new fields. Bigquery doesn't seem to support dropping columns within a record. One hacky is to
* Create new record column (details_new)
* Copy old record data into new record column without the expire column (details --> details_new)
* Delete old record column (details)
* Create a record column (empty details)
* Copy data into new record column (details_new --> details)

chowbao:
> Changes look good. Feel free to merge when ready.
> 
> While the raw data structure is named `ttl`, we might want to consider naming marts tables something a little more descriptive than ttl. 🤷‍♀️ that or we'll need to have robust docs so that end users know exactly what that means

Hmm we could rename it here in stellar-etl. Not sure

internal/transform/operation.go
sydneynotthecity: meh, I don't think it's worth the pain to do the drop and recopy (plus that would be pretty expensive) I already don't love the `history_operations` schema, and would prefer if we're making updates to the column to just convert to a `json` blob instead.

PR #210, opened 2023-12-12
Add soroban function symbol params

chowbao:
* Renamed export_expiration to export_ttl in the README
* Added human readable function names from soroban parameters
* Adds ledgerKey hash to contract_data and contract_code

### DIFF
--- a/internal/transform/contract_code.go
+++ b/internal/transform/contract_code.go
@@ -3,7 +3,6 @@ package transform
 import (
 	"fmt"
 
-	"github.com/stellar/go/hash"
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/strkey"
 	"github.com/stellar/go/xdr"
@@ -27,10 +26,7 @@ func TransformContractCode(ledgerChange ingest.Change, header xdr.LedgerHeaderHi
 		return ContractCodeOutput{}, nil
 	}
 
-	ledgerKey, _ := ledgerEntry.LedgerKey()
-	ledgerKeyByte, _ := ledgerKey.MarshalBinary()
-	hashedLedgerKeyByte := hash.Hash(ledgerKeyByte)
-	ledgerKeyHash, _ := strkey.Encode(strkey.VersionByteContract, hashedLedgerKeyByte[:])
+	ledgerKeyHash := utils.LedgerEntryToLedgerKeyHash(ledgerEntry)
 
 	contractCodeExtV := contractCode.Ext.V
 
--- a/internal/transform/contract_code.go
+++ b/internal/transform/contract_code.go
@@ -3,6 +3,7 @@ package transform
 import (
 	"fmt"
 
+	"github.com/stellar/go/hash"
 	"github.com/stellar/go/ingest"
 	"github.com/stellar/go/strkey"
 	"github.com/stellar/go/xdr"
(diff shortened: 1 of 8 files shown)


internal/transform/contract_code.go
```diff
@@ -26,6 +27,11 @@ func TransformContractCode(ledgerChange ingest.Change, header xdr.LedgerHeaderHi
 		return ContractCodeOutput{}, nil
 	}
 
+	ledgerKey, _ := ledgerEntry.LedgerKey()
+	ledgerKeyByte, _ := ledgerKey.MarshalBinary()
+	hashedLedgerKeyByte := hash.Hash(ledgerKeyByte)
+	ledgerKeyHash, _ := strkey.Encode(strkey.VersionByteContract, hashedLedgerKeyByte[:])
+
```
sydneynotthecity, line 34: Not a big deal, but might be interesting to encapsulate into a function if there are other instances in the future where we need to return `ledgerKeyHash`

internal/transform/operation.go
```diff
@@ -962,10 +963,14 @@ func extractOperationDetails(operation xdr.Operation, transaction ingest.LedgerT
...
 						serializedParam["value"] = base64.StdEncoding.EncodeToString(raw)
 					}
+					if scValTypeName == "Sym" {
+						decodedSymbolParamNames = append(decodedSymbolParamNames, param.String())
+					}
 				}
 				params = append(params, serializedParam)
 			}
 			details["parameters"] = params
+			details["decoded_symbol_param_names"] = decodedSymbolParamNames
```
sydneynotthecity, line 973: Just so I understand, we will be storing both the raw `scvSymbol` and the string version? Since the `decoded_symbol_param_names` are stored in a separate object from `parameters`, will there be a way to associate which raw `scvSymbol` = the decoded name?

sydneynotthecity (approved):
LGTM

internal/transform/operation.go
sreuland: is `decoded_symbol_param_names` used elsewhere in etl, I didn't see it ref'd in https://github.com/stellar/stellar-etl-airflow/pull/270, is it intended to be persisted as-is, and meant as a convenience for end clients to reference from sql instead of resolving at sql query time with a select `CASE WHEN parameters.type=Sym THEN decode(parameters.value, 'base64') ELSE ''`

or to maintain the association with the param order, could add it to `serializedParam["sym_str_val"]`

internal/transform/contract_code.go
chowbao: Oh yeah I should make this a function

internal/transform/operation.go
chowbao: >Just so I understand, we will be storing both the raw scvSymbol and the string version?

Yes

>will there be a way to associate which raw scvSymbol = the decoded name

Err kind of. You would need to decode the scvSymbol and then know that it matched the decoded name. Is there a reason why we would want this though? Even if it was easy to reference back between the two I don't think any useful information would be gained (scvSymbol param value is basically the decoded name).

chowbao: >is it intended to be persisted as-is

Yes I intended it to be used as-is. In BigQuery the user would query for the direct specific `decoded_symbol_param_names` (order shouldn't matter)

But depending on the Syd's response to my question above, maybe this will change

sydneynotthecity: > Is there a reason why we would want this though?

@chowbao no. My thought was if we ever needed to extract the _value_ submitted to a function call, it would be hard to trace back to the raw scvSymbol, but as @sreuland pointed out, we can actually just perform the transformation ourselves in SQL. Nice catch! I didn't realize these values were just b64 encoded.

> is `decoded_symbol_param_names` used elsewhere in etl, I didn't see it ref'd

The new field is ref'd [here](https://github.com/stellar/stellar-etl-airflow/pull/270/files#diff-8d121f1d0361692b8073f42be06442a97efd82499e153a3f34251f67f7ddaedcR1060-R1064) in the PR. 

Really the only reason to decode is to help with usability, we could leave it in the raw format and decode in downstream tables, but there is also no compelling reason to leave b64 encoded.

chowbao: > if we ever needed to extract the value submitted to a function call

We could add a serialized decoded parameters then? scvSymbol wouldn't have a value aside from the function name so you'd need to know the next relevant parameter values. And if that's the case we could just provide 1 level of decoding (no nested parameter values). Which would then become like the following:

```
serializedParamDecoded := map[string]string{}
serializedParamDecoded["type"] = scValTypeName
serializedParamDecoded["value"] = param.String()

details["params_decoded"] = paramsDecoded
```

PR #211, opened 2023-12-18
Add fee bump envelope to soroban operations

chowbao:
There is a missing TransactionEnvelope type that was added `EnvelopeTypeEnvelopeTypeTxFeeBump`
Added a function to decode the TransactionEnvelope correctly to return the `contract_id` and `contract_code_hash`


internal/transform/operation.go
```diff
@@ -1639,6 +1639,17 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
 	return details, nil
 }
 
+func getTransactionV1Envelope(transactionEnvelope xdr.TransactionEnvelope) xdr.TransactionV1Envelope {
+	switch transactionEnvelope.Type {
+	case xdr.EnvelopeTypeEnvelopeTypeTx:
+		return transactionEnvelope.MustV1()
+	case xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
+		return transactionEnvelope.MustFeeBump().Tx.InnerTx.MustV1()
```
sydneynotthecity, line 1647: Interesting that we have to switch this ourselves. I'm kinda surprised there isn't an function in the xdr already to do this.

TIL there are more EnvelopeTypes! Do you think we need to handle other envelope types like `EnvelopeTypeEnvelopeTypeContractId ` or `EnvelopeTypeEnvelopeTypePoolRevokeOpId `?

sydneynotthecity (approved):
LGTM

PR #213, opened 2023-12-19
Change configs from expiration to ttl

chowbao:
Change export type flags from `expiration` to `ttl` to fix missing ttl ledger entries


sydneynotthecity (approved):
LGTM

PR #215, opened 2024-01-09
Add common interface to write exported files

chowbao:
* Added a common interface to write exported files to a chosen cloud provider storage service.
* Moved utility functions out of export_ledgers
* Added gcs to the common interface

Note: Because of the way the cmd package was written the `command_utils` and `upload_to_gcs` are at the same cmd/ directory level in order to avoid more refactoring. This is mostly due to cmdLogger (from root.go) and being embedded in a bunch of functions throughout the cmd package


internal/utils/main.go
```diff
@@ -337,16 +338,23 @@ func MustBucketFlags(flags *pflag.FlagSet, logger *EtlLogger) (path string) {
 	return
 }
 
-// MustGcsFlags gets the values of the bucket list specific flags: gcp-project and gcs-bucket
-func MustGcsFlags(flags *pflag.FlagSet, logger *EtlLogger) (bucket, credentials string) {
-	bucket, err := flags.GetString("gcs-bucket")
+// MustCloudStorageFlags gets the values of the bucket list specific flags: cloud-storage-bucket, cloud-credentials
+func MustCloudStorageFlags(flags *pflag.FlagSet, logger *EtlLogger) (bucket, credentials, provider string) {
```
sydneynotthecity, line 342: By making this method required, if a user only wants to read/write the files locally and deal with transmission in another process, are they going to get a warning every time they export without specifying a storage destination?

cmd/export_orderbooks.go
sydneynotthecity, line 1: At some point we should evaluate whether it makes sense to deprecate this entirely.

sydneynotthecity (approved):
LGTM

internal/utils/main.go
sydneynotthecity: I almost wonder if we make writing to a data store something you opt into, like the `--testnet` parameter and then pass accompanying cloud provider flags

sydneynotthecity: > Well the funny thing is the precedent in the code is to make it required but give it a default value 😆

yeah, yeah i know 😅 

i like that idea, i'm fine incrementally improving this interface so we can clean it up once we refactor stellar-etl to use a different ledgerbackend interface (ie, reading from GCS)

PR #219, opened 2024-01-16
Add user agent to captive core config

chowbao:
* Add missing comma to export_effets.go
* Re-add `https://history.stellar.org/prd/core-live/core_live_003` for history archive URL list
* Choose random history archive url


internal/utils/main.go
```diff
@@ -538,7 +539,15 @@ var futureArchiveURLs = []string{
...
-	return historyarchive.NewArchivePool(archiveURLS, historyarchive.ConnectOptions{})
+	var archiveInterface historyarchive.ArchiveInterface
+
+	archivePool, err := historyarchive.NewArchivePool(archiveURLS, historyarchive.ConnectOptions{UserAgent: "stellar-etl/1.0.0"})
+	if err != nil {
+		return archiveInterface, err
+	}
+
+	// Use random archive pool
+	return archivePool[int32(rand.Intn(len(archivePool)))], nil
```
Shaptic, line 550: I think you should just be returning the pool itself and let it randomize on every invocation rather than picking one that’s fixed randomly.

chowbao: So just return ArchivePool.
I'm not sure how it returns a random instance though. Is that a golang thing? My guess was it was always sending the first element in the array.

sydneynotthecity (approved):
LGTM

internal/utils/main.go
sydneynotthecity: Yeah, it's a golang thing when the [NewArchivePool](https://github.com/stellar/go/blob/master/historyarchive/archive_pool.go) is created. The code actually connects to all HAS URLs and then will randomly select one of the connections when a method is called

Shaptic: Yep, exactly. Basically `ArchivePool` acts **exactly** like a normal archive by providing the same interface, but under the hood it's selecting a random archive.

PR #220, opened 2024-01-18
Add export all history command

chowbao:
Adding temporary command `export_all_history` to mitigate the high egress costs hitting the history archives.
This combines the export for operations, trades, effects, transactions, and diagnostic events to use only a single captive core instead of each spinning up their own for data export.

TODO: This should be removed as soon as LedgerExporter is complete and a TxMeta datalake is available

### DIFF
--- a/cmd/export_all_history.go
+++ b/cmd/export_all_history.go
@@ -22,7 +22,7 @@ in order to mitigate egress costs for the entity hosting history archives.`,
 		cmdLogger.SetLevel(logrus.InfoLevel)
 		endNum, strictExport, isTest, isFuture, extra := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
 		cmdLogger.StrictExport = strictExport
-		startNum, _, limit := utils.MustArchiveFlags(cmd.Flags(), cmdLogger)
+		startNum, path, limit := utils.MustArchiveFlags(cmd.Flags(), cmdLogger)
 		gcsBucket, gcpCredentials := utils.MustGcsFlags(cmd.Flags(), cmdLogger)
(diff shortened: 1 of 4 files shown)


cmd/export_all_history.go
```diff
@@ -0,0 +1,207 @@
...
+		if err != nil {
+			cmdLogger.LogError(fmt.Errorf("could not export operation: %v", err))
+			numFailures += 1
+			continue
+		}
+		totalNumBytes += numBytes
+	}
+
+	outFileOperations.Close()
+	cmdLogger.Info("Number of bytes written: ", totalNumBytes)
```
sydneynotthecity, line 67: If we get frequent transient errors, more logging that identifies which file/export process failed might be helpful. Since this is a temp fix, i'll leave it up to you whether that's worthwhile to add.

sydneynotthecity (commented):
Thanks for the speedy turnaround with the changes. Looks like a good enough temporary fix, so I kept the comments to a minimum.

How does the runtime for all `history_*` exports compare with running a single export command?

cmd/export_all_history.go
chowbao: Yeah it's probably not that useful. Any issue would more easily be resolved by running the original pipeline instead of this hack

chowbao:
> How does the runtime for all `history_*` exports compare with running a single export command?

I didn't run this in prod but in test it adds about 10s - 20s more time (so completes in 40s compared to 20s-30s). I _think_ the runtime ratio will be much better in prod because the part that will benefit the most is the captive core catchup portion

sydneynotthecity (approved):

PR #221, opened 2024-01-29
use NewLedgerChangeReaderFromLedgerCloseMeta

sfsf9797:
would it be more efficient to use NewLedgerTransactionReaderFromLedgerCloseMeta?  feel free to close this if the PR doesn't make sense. Thanks.

### DIFF
--- a/internal/input/transactions.go
+++ b/internal/input/transactions.go
@@ -38,7 +38,7 @@ func GetTransactions(start, end uint32, limit int64, env utils.EnvironmentDetail
 			return nil, errors.Wrap(err, "error getting ledger from the backend")
 		}
 
-		txReader, err := ingest.NewLedgerChangeReaderFromLedgerCloseMeta(env.NetworkPassphrase, ledgerCloseMeta)
+		txReader, err := ingest.NewLedgerTransactionReaderFromLedgerCloseMeta(env.NetworkPassphrase, ledgerCloseMeta)
 		if err != nil {
 			return []LedgerTransformInput{}, err
 		}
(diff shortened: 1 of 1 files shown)


internal/input/transactions.go
```diff
@@ -33,17 +33,18 @@ func GetTransactions(start, end uint32, limit int64, env utils.EnvironmentDetail
 	err = backend.PrepareRange(ctx, ledgerbackend.BoundedRange(start, end))
 	panicIf(err)
 	for seq := start; seq <= end; seq++ {
-		txReader, err := ingest.NewLedgerTransactionReader(ctx, backend, env.NetworkPassphrase, seq)
```
chowbao, line 36: Hmm actually the returned LedgerTransactionReader (the txReader in this case) has ledgerCloseMeta inside it. So the refactor should be to 
* remove the extra ledgerCloseMeta in line 43
* update LedgerCloseMeta in line 57 to be set by `txReader.ledgerCloseMeta`

sfsf9797: seem like txReader.ledgerCloseMeta field is not exported?

 https://github.com/stellar/go/blob/master/ingest/ledger_transaction_reader.go#L17

chowbao: I believe it should be
`NewLedgerTransactionReader()` calls `NewLedgerTransactionReaderFromLedgerCloseMeta()` which then sets  `ledgerCloseMeta` [here](https://github.com/stellar/go/blob/master/ingest/ledger_transaction_reader.go#L36)

sfsf9797: oh sorry, I might have misunderstood it. 
what I am trying to say is the ledgerCloseMeta field is not exported as it is not in capital letters. 


image

chowbao: Ah I missed that. You're right then

chowbao (approved):

PR #224, opened 2024-02-16
Delete id from hist_assets and change id to asset_id in claimable_balances and offer

cayod:
To unify the ID fields within the `history_assets` table, this PR implements the following adjustments:

- The `id` field has been removed from the `history_assets` table.

- The fields `asset_id` in `claimable_balance`, as well as `SellingAssetID` and `BuyingAssetID` in `offer`, now utilize `asset_id` for identification purposes instead of `id`.


cmd/export_assets.go
```diff
@@ -29,7 +29,7 @@ var assetsCmd = &cobra.Command{
 		}
 
 		// With seenIDs, the code doesn't export duplicate assets within a single export. Note that across exports, assets may be duplicated
-		seenIDs := map[uint64]bool{}
+		seenIDs := map[int64]bool{}
```
sydneynotthecity, line 32: Doesn't this need to stay `uint64`?

cayod: @sydneynotthecity It is because in the `FarmHashAsset` function we change the uint64 to int64, so asset_id is an int64 as was discussed in this [PR](https://github.com/stellar/stellar-etl/pull/183#issuecomment-1478356546)

sydneynotthecity (approved):
LGTM

PR #227, opened 2024-02-22
Fix soroban hashes to hexstring

chowbao:
XDR hash should be printed as hexstring instead of strkey(versionByteContract)


sydneynotthecity (approved):
LGTM

PR #228, opened 2024-03-05
Log Transaction Codes and Operation Traces

sydneynotthecity:
Adding transaction and operation level trace codes to stellar-etl. This gives a more detailed view into error reasons when a transaction fails to execute successfully.

### DIFF
--- a/internal/transform/operation.go
+++ b/internal/transform/operation.go
@@ -66,15 +66,28 @@ func TransformOperation(operation xdr.Operation, operationIndex int32, transacti
 		return OperationOutput{}, err
 	}
 
+	outputOperationResults, ok := transaction.Result.Result.OperationResults()
+	if !ok {
+		return OperationOutput{}, err
+	}
+	outputOperationResultCode := outputOperationResults[operationIndex].Code.String()
+	// TODO: add trace code details
+	outputOperationTraceCode, err := mapOperationTrace(*outputOperationResults[operationIndex].Tr)
+	if err != nil {
+		return OperationOutput{}, err
+	}
+
 	transformedOperation := OperationOutput{
-		SourceAccount:      outputSourceAccount,
-		SourceAccountMuxed: outputSourceAccountMuxed.String,
-		Type:               outputOperationType,
-		TypeString:         outputOperationTypeString,
-		TransactionID:      outputTransactionID,
-		OperationID:        outputOperationID,
-		OperationDetails:   outputDetails,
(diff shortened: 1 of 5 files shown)


internal/transform/operation.go
```diff
@@ -145,6 +157,71 @@ func mapOperationType(operation xdr.Operation) (string, error) {
...
+	case xdr.OperationTypeLiquidityPoolWithdraw:
+		operationTraceDescription = operationTrace.LiquidityPoolWithdrawResult.Code.String()
+	case xdr.OperationTypeInvokeHostFunction:
+		operationTraceDescription = operationTrace.InvokeHostFunctionResult.Code.String()
+	case xdr.OperationTypeExtendFootprintTtl:
+		operationTraceDescription = operationTrace.ExtendFootprintTtlResult.Code.String()
+	case xdr.OperationTypeRestoreFootprint:
+		operationTraceDescription = operationTrace.RestoreFootprintResult.Code.String()
+	default:
+		return operationTraceDescription, fmt.Errorf("Unknown operation type: %s", operationTrace.Type.String())
```
chowbao, line 220: I think `operationTrace.Type.String()` would cause an error if the OperationType is bad.

```
// String returns the name of `e`
func (e OperationType) String() string {
	name, _ := operationTypeMap[int32(e)]
	return name
}
```
And `operationTypeMap[int32(e)]` would return a invalid key in map or something

sydneynotthecity: I thought this default error was designed to raise when the Operation Type is _valid_, we just haven't mapped it in our case statements. The behavior mirrors the way we define [mapOperationType](https://github.com/stellar/stellar-etl/blob/master/internal/transform/operation.go#L83). My understanding is if the operationtype itself is bad, our code would already error and be unable to parse that operation.

sydneynotthecity: Do you want to change how we handle that in both functions? And if so, any guidance? Looks like the `operation.Type.String()` function does not return an error, so we don't have the ability to pass the error through

chowbao: > My understanding is if the operationtype itself is bad, our code would already error and be unable to parse that operation

Ah you're right. I think the only time this would actually be a problem is if XDR was updated ahead of what stellar-etl is using. And most likely something else would error out before it even reached this point

> Do you want to change how we handle that in both functions?

Nope it's fine as is

chowbao (approved):

PR #229, opened 2024-03-06
Enable diagnostic events

chowbao:
* Enable diagnostic events in stellar core configs
* Use GetDiagnosticEvents() which converts contract events to diagnostic events in the absence of diagnostic events


sydneynotthecity (approved):
LGTM

PR #231, opened 2024-03-11
Update soroban fees

chowbao:
* Added `inclusion_fee_bid, inclusion_fee_charged, and resource_fee_refund`
* Updated fee_charged to be the correct amount (current fee_charged value from core is incorrect due to a bug in protocol 20)

### DIFF
--- a/internal/transform/transaction.go
+++ b/internal/transform/transaction.go
@@ -138,6 +138,9 @@ func TransformTransaction(transaction ingest.LedgerTransaction, lhe xdr.LedgerHe
 	var outputSorobanResourcesInstructions uint32
 	var outputSorobanResourcesReadBytes uint32
 	var outputSorobanResourcesWriteBytes uint32
+	var outputInclusionFeeBid int64
+	var outputInclusionFeeCharged int64
+	var outputResourceFeeRefund int64
 
 	transactionEnvelopeV1, ok := transaction.Envelope.GetV1()
 	if ok {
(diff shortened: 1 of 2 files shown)


internal/transform/transaction.go
```diff
@@ -147,6 +150,20 @@ func TransformTransaction(transaction ingest.LedgerTransaction, lhe xdr.LedgerHe
...
+			outputInclusionFeeCharged = initialFeeCharged - outputResourceFee
+
+			meta, ok := transaction.UnsafeMeta.GetV3()
+			if ok {
+				accountBalanceStart, accountBalanceEnd := getAccountBalanceFromLedgerEntryChanges(meta.TxChangesAfter, sourceAccount.Address())
+				outputResourceFeeRefund = accountBalanceEnd - accountBalanceStart
+			}
+
+			// TODO: FeeCharged is calculated incorrectly in protocol 20. Remove when protocol is updated and the bug is fixed
+			outputFeeCharged = outputFeeCharged - outputResourceFeeRefund
```
sydneynotthecity, line 166: Just confirming that this won't have a negative impact on classic transactions since the `outputResourceFeeRefund` should always be 0 for classic, correct?

internal/transform/transaction.go
```diff
@@ -216,6 +236,36 @@ func TransformTransaction(transaction ingest.LedgerTransaction, lhe xdr.LedgerHe
...
+				accountBalanceEnd = int64(accountEntry.Balance)
+			}
+		case xdr.LedgerEntryChangeTypeLedgerEntryState:
+			accountEntry, ok := change.State.Data.GetAccount()
+			if !ok {
+				continue
+			}
+
+			if accountEntry.AccountId.Address() == sourceAccountAddress {
+				accountBalanceStart = int64(accountEntry.Balance)
```
sydneynotthecity, line 261: will this logic work with fee bump transactions as well?

sydneynotthecity (commented):
Changes look good, I have a question about fee bump transactions and will approve once answered

internal/transform/transaction.go
chowbao: Yeah this should work for classic. This is wrapped within a soroban only check for everything in this code block

sydneynotthecity: ah i missed that part, thanks

internal/transform/transaction.go
chowbao: None of the above logic runs for a fee bump transaction. Right now it only looks at V1 transactions which can have [sorobanData](https://github.com/stellar/stellar-etl/pull/231/files/213cb28a5564ad5b0d02f48d2aae09ffef3b043b#diff-c26dd0484eb96ce222a4019ab0809b202fd4b82ded179f0806674123667fd419R145)

Right now the fee for fee bump transactions is saved in `max_fee`. Do these fees need some breakdown as well?

chowbao: Updated to handle fee bump transactions as well. Although there aren't any in mainnet nor testnet so far

sydneynotthecity (approved):
LGTM

PR #233, opened 2024-03-19
Release drafter and publisher

laysabit:
Everytime there is a new pushed tag in any branch in the repository, the process of publishing the release based on that tag starts. That way, the tag can be used to create a release, and also add new functionalities and patches.

It was designed based on semantic versioning: major.minor.patches . v1.0.0, v1.0.1, v0.1.1.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
       - uses: release-drafter/release-drafter@v5
         with:
           commitish: master
-          name: "stellar-dbt ${{ steps.gettag.outputs.TAG }}"
+          name: "stellar-etl ${{ steps.gettag.outputs.TAG }}"
           tag: ${{ github.ref }}
           publish: true
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      # ${{ github.ref }} was not giving v* as tag name, but refs/tags/v* instead, so I had to abbreviate it
+      - name: Get latest abbreviated tag
+        id: gettag
+        run: echo ::set-output name=TAG::$(git describe --tags --abbrev=7)
+
+      - uses: release-drafter/release-drafter@v5
+        with:
+          commitish: master
+          name: "stellar-dbt ${{ steps.gettag.outputs.TAG }}"
```
sydneynotthecity, line 33: I think this need to be `stellar-etl`

laysabit: Done

chowbao (approved):

PR #234, opened 2024-03-21
Fix fee charged calculation

chowbao:
Fee charged should actually be `resource fee - resource fee refund + inclusion fee charged`


sydneynotthecity (approved):
LGTM

PR #237, opened 2024-04-11
Add `ledger_key_hash` to `history_operations`

lucaszanotelli:
This PR introduces a new field named `ledger_key_hash` to the `history_operations` export file. The field will only be included when the OperationType property is one of the following values:

- `InvokeHostFunction`
- `ExtendFootprintTtl`
- `RestoreFootprint`

### DIFF
--- a/internal/transform/operation.go
+++ b/internal/transform/operation.go
@@ -1792,22 +1792,21 @@ func contractCodeHashFromTxEnvelope(transactionEnvelope xdr.TransactionV1Envelop
 	return ""
 }
 
-func ledgerKeyHashFromTxEnvelope(transactionEnvelope xdr.TransactionV1Envelope) string {
+func ledgerKeyHashFromTxEnvelope(transactionEnvelope xdr.TransactionV1Envelope) []string {
+	var ledgerKeyHash []string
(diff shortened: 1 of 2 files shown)


internal/transform/operation.go
```diff
@@ -1782,6 +1792,24 @@ func contractCodeHashFromTxEnvelope(transactionEnvelope xdr.TransactionV1Envelop
 	return ""
 }
 
+func ledgerKeyHashFromTxEnvelope(transactionEnvelope xdr.TransactionV1Envelope) string {
+	for _, ledgerKey := range transactionEnvelope.Tx.Ext.SorobanData.Resources.Footprint.ReadOnly {
+		ledgerKeyHash := utils.LedgerKeyToLedgerKeyHash(ledgerKey)
+		if ledgerKeyHash != "" {
+			return ledgerKeyHash
+		}
+	}
```
sydneynotthecity, line 1801: we just learned that restore/bump footprint ops will bump all keys in the footprint. Is it possible to store all keys in an array instead of saving only the first one?

sydneynotthecity (commented):
Logic is correct; it just needs to be extended to cover all keys that are bumped/restored. The footprint can contain more than one key, so we will need to list all of them.

sydneynotthecity (approved):
LGTM

PR #238, opened 2024-04-18
Feature/change workdir

edualvess:
### Context
This PR aims to set the container `WORKDIR` on the image to a path not used in building stages that could be later mounted using ephemeral disks. The former path used is not suitable for the ephemeral volume mount mentioned since the building process persists data on it, and mounting an empty volume on this path would cause the container to lose the data required.
This change also requires the Stellar core configuration path to be adjusted for commands where the CoreConfig is not passed as an argument since it is gotten from environment details and later set to paths relative to the container `WORKDIR`. 

### Key Changes
- Changed docker `WORKDIR` to `/etl/data` from `/etl`
- Set the utils EnvironmentDetails values for `CoreConfig` to refer to the absolute path instead of the relative path.

### DIFF
--- a/cmd/export_ledger_entry_changes_test.go
+++ b/cmd/export_ledger_entry_changes_test.go
@@ -6,7 +6,7 @@ import (
 )
 
 const coreExecutablePath = "../stellar-core/src/stellar-core"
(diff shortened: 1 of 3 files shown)


sydneynotthecity (approved):
LGTM

PR #240, opened 2024-04-29
Add fee fields; formatting for warnings

chowbao:
* Add new soroban fee fields
* input/ledgers needed to add LCM to extract a fee field
  * The fee field will be blank if running in captive-core mode because HistoryArchives does not contain the value
  * Potentially can add to the captive-core mode but doesn't seem worth the effort as everything runs with txmeta files now
* Fixed all existing warnings
  * Warnings were mostly from fmt.errors should not start with a capitalized string
  * or unused functions/variables

Note: The `EMIT_SOROBAN_TRANSACTION_META_EXT_V1` `EMIT_LEDGER_CLOSE_META_EXT_V1` flags for core are set to true for ledgerexporter testnet txmeta files


internal/input/ledgers.go
```diff
@@ -68,7 +68,12 @@ func GetLedgers(start, end uint32, limit int64, env utils.EnvironmentDetails, us
 			},
 		}
 
-		ledgerSlice = append(ledgerSlice, ledger)
+		ledgerLCM := utils.HistoryArchiveLedgerAndLCM{
```
chowbao, line 71: Because of the new LCM fee field we need to pass in the `historyarchive.Ledger` as well as the `LedgerCloseMeta`. This should be eventually refactored in the future to where `input/ledger` and `input/assets` don't use `historyarchive.Ledger` and just use LCM

internal/transform/transaction.go
```diff
@@ -167,6 +167,12 @@ func TransformTransaction(transaction ingest.LedgerTransaction, lhe xdr.LedgerHe
 		if ok {
 			accountBalanceStart, accountBalanceEnd := getAccountBalanceFromLedgerEntryChanges(meta.TxChangesAfter, sourceAccount.Address())
 			outputResourceFeeRefund = accountBalanceEnd - accountBalanceStart
+			extV1, ok := meta.SorobanMeta.Ext.GetV1()
+			if ok {
+				outputTotalNonRefundableResourceFeeCharged = int64(extV1.TotalNonRefundableResourceFeeCharged)
+				outputTotalRefundableResourceFeeCharged = int64(extV1.TotalRefundableResourceFeeCharged)
+				outputRentFeeCharged = int64(extV1.RentFeeCharged)
+			}
```
chowbao, line 175: New fee fields

internal/transform/ledger.go
```diff
@@ -55,6 +55,12 @@ func TransformLedger(inputLedger historyarchive.Ledger) (LedgerOutput, error) {
 
 	outputProtocolVersion := uint32(ledgerHeader.LedgerVersion)
 
+	var outputSorobanFeeWrite1Kb int64
+	lcmV1, ok := lcm.GetV1()
+	if ok {
+		outputSorobanFeeWrite1Kb = int64(lcmV1.Ext.V1.SorobanFeeWrite1Kb)
```
chowbao, line 61: New fee change

sreuland: may be worthwhile to use the safety getter method of `LedgerCloseMetaExt.GetV1()` rather than direct reference to nested `V1` struct, to avoid potential nil pointer,  i.e. check `ok` from `lcmv1.Ext.GetV1()`

chowbao: Makes sense. Done in  [d43e728](https://github.com/stellar/stellar-etl/pull/240/commits/d43e728fe285fd51bd2cd30fb18e5c6634bff6a6)

internal/transform/contract_data.go
```diff
@@ -10,28 +10,28 @@ import (
...
-	adminSym           = xdr.ScSymbol("Admin")
-	issuerSym          = xdr.ScSymbol("issuer")
-	assetCodeSym       = xdr.ScSymbol("asset_code")
-	alphaNum4Sym       = xdr.ScSymbol("AlphaNum4")
-	alphaNum12Sym      = xdr.ScSymbol("AlphaNum12")
-	decimalSym         = xdr.ScSymbol("decimal")
-	assetInfoSym       = xdr.ScSymbol("AssetInfo")
-	decimalVal         = xdr.Uint32(scDecimalPrecision)
-	assetInfoVec       = &xdr.ScVec{
+	//metadataSym        = xdr.ScSymbol("METADATA")
```
sydneynotthecity, line 23: Should these commented fields just be removed from the codebase? Were they not emitted in P20 upgrade?

internal/transform/effects_test.go
```diff
@@ -3326,28 +3327,28 @@ func getRevokeSponsorshipMeta(t *testing.T) (string, []EffectOutput) {
...
-	order        uint32
-}
+//type effect struct {
+//	address      string
+//	addressMuxed null.String
+//	operationID  int64
+//	details      map[string]interface{}
+//	effectType   EffectType
+//	order        uint32
+//}
```
sydneynotthecity, line 3351: Are these needed at all?

internal/transform/ledger_test.go
```diff
@@ -88,11 +114,13 @@ func makeLedgerTestOutput() (output LedgerOutput, err error) {
 		SuccessfulTransactionCount: 1,
 		FailedTransactionCount:     1,
 		TxSetOperationCount:        "13",
+
```
sydneynotthecity, line 117: nit: extra return

internal/transform/schema.go
```diff
@@ -28,46 +28,50 @@ type LedgerOutput struct {
...
+	ResourceFee                          int64          `json:"resource_fee"`
+	SorobanResourcesInstructions         uint32         `json:"soroban_resources_instructions"`
+	SorobanResourcesReadBytes            uint32         `json:"soroban_resources_read_bytes"`
+	SorobanResourcesWriteBytes           uint32         `json:"soroban_resources_write_bytes"`
+	TransactionResultCode                string         `json:"transaction_result_code"`
+	InclusionFeeBid                      int64          `json:"inclusion_fee_bid"`
+	InclusionFeeCharged                  int64          `json:"inclusion_fee_charged"`
+	ResourceFeeRefund                    int64          `json:"resource_fee_refund"`
+	TotalNonRefundableResourceFeeCharged int64          `json:"non_refundable_resource_fee_charged"`
+	TotalRefundableResourceFeeCharged    int64          `json:"refundable_resource_fee_charged"`
```
sydneynotthecity, line 73: do these differ from `resource_fee_refund` and `resource_fee`?

sydneynotthecity (commented):
The `HistoryArchiveLedgerAndLCM` is definitely a hack that I want to address and refactor sometime this quarter. Probably a good opportunity to pair program with a new team member and refactor that code. Like you noted, if someone runs the codebase using Captive Core, they will run into errors getting the write fee since LCM will be empty.

Also, do you think there is any merit in keeping the code you commented out, or should that just be deleted entirely? A lot of those functions predate me and were original attempts to maintain parity between Horizon/Hubble.

sydneynotthecity:
> Note: The EMIT_SOROBAN_TRANSACTION_META_EXT_V1 EMIT_LEDGER_CLOSE_META_EXT_V1 flags for core are set to true for ledgerexporter testnet txmeta files

Nit: Should we also update the Captive core configs to use those flags within `stellar-etl` too?

internal/transform/contract_data.go
chowbao: They can be removed. These were defined for [unit tests for Horizon](https://github.com/stellar/go/blob/master/services/horizon/internal/ingest/processors/contract_data.go#L416)

internal/transform/schema.go
chowbao: From the xdr itself they are different fields. Functionally equivalent I'm not sure. We can probably compare them when there is data emitted and then determine if we only need to keep one or the other

chowbao:
>Also, do you think there is any merit in keeping the code you commented out, or should that just be deleted entirely?

Yeah we can delete them. Worse case we can find them again in the commit history

internal/transform/effects_test.go
chowbao: Probably not. Removing

sydneynotthecity (approved):
LGTM

PR #241, opened 2024-04-30
Add contract code fees

chowbao:
Contract code resource breakdown was added in [CAP-0054](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0054.md). Pull in these new fields for contract code ledgers

### DIFF
--- a/internal/transform/contract_code.go
+++ b/internal/transform/contract_code.go
@@ -38,6 +38,31 @@ func TransformContractCode(ledgerChange ingest.Change, header xdr.LedgerHeaderHi
 
 	ledgerSequence := header.Header.LedgerSeq
 
+	var outputNInstructions uint32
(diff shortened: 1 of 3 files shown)


chowbao:
> From the CAP, it also looks like there were fields added to the `ConfigSettings` ledger entry. Is that true? Do you think they're worth adding to `config_settings`?

There were no config_setting changes. It was additional fields to `ContractCostType` which isn't used in config_settings

sydneynotthecity:
> There were no config_setting changes. It was additional fields to ContractCostType which isn't used in config_settings

Ah, I saw `Stellar-contract-config-setting.x` and assumed it was saved on `config_settings`. I see this is just the enum

sydneynotthecity (approved):

PR #242, opened 2024-05-07
Update to use BufferedStorageBackend to read txmeta files

chowbao:
Replaces `GCSBackend` with the released `BufferedStorageBackend` ledgerbackend reader in the go monorepo

* Update parameters needed for `BufferedStorageBackend`
* Save common parameters in a struct instead of passing individual values
  * This should be the case for all parameters; TODO in a separate PR
* The param `datastore-url` is now `datastore-path` as the backend only takes the path
* `Network` added to env vars


cmd/export_account_signers.go
```diff
@@ -22,13 +22,13 @@ should be used in an initial data dump. In order to get account information with
 the export_ledger_entry_changes command.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		cmdLogger.SetLevel(logrus.InfoLevel)
-		endNum, strictExport, isTest, isFuture, extra, _, datastoreUrl := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
-		cmdLogger.StrictExport = strictExport
-		env := utils.GetEnvironmentDetails(isTest, isFuture, datastoreUrl)
+		commonArgs := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
+		cmdLogger.StrictExport = commonArgs.StrictExport
+		env := utils.GetEnvironmentDetails(commonArgs.IsTest, commonArgs.IsFuture, commonArgs.DatastorePath)
```
sydneynotthecity: not blocking for this PR: wdty about passing an env arg instead of `IsTest` and `IsFuture`? I think we'd need to plan the breaking change with Goldsky, but might be a better UX

cmd/export_assets.go
```diff
@@ -16,21 +16,21 @@ var assetsCmd = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
 		cmdLogger.SetLevel(logrus.InfoLevel)
-		endNum, strictExport, isTest, isFuture, extra, useCaptiveCore, datastoreUrl := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
-		cmdLogger.StrictExport = strictExport
+		commonArgs := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
+		cmdLogger.StrictExport = commonArgs.StrictExport
 		startNum, path, limit := utils.MustArchiveFlags(cmd.Flags(), cmdLogger)
 		cloudStorageBucket, cloudCredentials, cloudProvider := utils.MustCloudStorageFlags(cmd.Flags(), cmdLogger)
-		env := utils.GetEnvironmentDetails(isTest, isFuture, datastoreUrl)
+		env := utils.GetEnvironmentDetails(commonArgs.IsTest, commonArgs.IsFuture, commonArgs.DatastorePath)
```
sydneynotthecity, line 23: Not blocking the PR, but wdty about passing an environment arg in the future? I think it would be a better UX instead of `IsTest` and `IsFuture`. We would need to coordinate with Goldsky if they're using our ETL

cmd/export_ledger_transaction.go
sydneynotthecity, line 1: Todo: This export process may be less useful now that we save raw XDR to GCS

internal/utils/main.go
```diff
@@ -757,28 +795,42 @@ func CreateLedgerBackend(ctx context.Context, useCaptiveCore bool, env Environme
...
+	// TODO: In the future these will come from a config file written by ledgerexporter
+	// Hard code ledger batch values for now
+	ledgerBatchConfig := datastore.LedgerBatchConfig{
+		LedgersPerFile:    1,
+		FilesPerPartition: 64000,
+		FileSuffix:        ".xdr.gz",
+	}
+
+	// TODO: In the future CompressionType should be removed as it won't be configurable
+	BSBackendConfig := ledgerbackend.BufferedStorageBackendConfig{
```
sydneynotthecity, line 819: lol, BSBackendConfig :eyeroll:

internal/utils/main.go
```diff
@@ -757,28 +795,42 @@ func CreateLedgerBackend(ctx context.Context, useCaptiveCore bool, env Environme
...
+		LedgerBatchConfig: ledgerBatchConfig,
+		CompressionType:   "gzip",
+		DataStore:         dataStore,
+		BufferSize:        1000,
+		NumWorkers:        5,
+		RetryLimit:        3,
+		RetryWait:         5,
```
sydneynotthecity, line 826: If these values are hardcoded here, why are we passing in as flags?

internal/utils/main.go
```diff
@@ -757,28 +795,42 @@ func CreateLedgerBackend(ctx context.Context, useCaptiveCore bool, env Environme
...
-	fileConfig := ledgerbackend.LCMFileConfig{
-		StorageURL:        env.StorageURL,
-		FileSuffix:        ".xdr.gz",
-		LedgersPerFile:    1,
-		FilesPerPartition: 64000,
+	params := make(map[string]string)
+	params["destination_bucket_path"] = env.DatastorePath
+	dataStoreConfig := datastore.DataStoreConfig{
+		Type:   "GCS",
+		Params: params,
```
sydneynotthecity, line 802: Are there any other optional params allowed in `DataStoreConfig`? I'm just curious

sydneynotthecity (approved):
Nothing blocking the merge. Changes look good

Might be worthwhile for us to discuss the operator experience for stellar-etl since we're also discussing for Ledger Exporter. Having the flags called through multiple methods feels weird now that we're deleting our extract from genesis code. Specifically, it might be worthwhile for us to collapse `MustCommonFlags` and `MustArchiveFlags`

cmd/export_account_signers.go
chowbao: Yeah that'd be good. We need to refactor a lot of the params TBH 

[Hubble-386](https://stellarorg.atlassian.net/jira/software/c/projects/HUBBLE/boards/65/backlog?selectedIssue=HUBBLE-386) although it doesn't have a description lol

cmd/export_assets.go
chowbao: Yes agreed. 

Resolve because this comment is a duplicate of https://github.com/stellar/stellar-etl/pull/242#discussion_r1593091209

internal/utils/main.go
chowbao: That's because I forgot to update the code to use the new flags 😆

internal/utils/main.go
chowbao: Not currently

cmd/export_ledger_transaction.go
chowbao: Hmm that's true. Throwing it in the ticket lol
[Hubble-386](https://stellarorg.atlassian.net/jira/software/c/projects/HUBBLE/boards/65/backlog?selectedIssue=HUBBLE-386)

PR #245, opened 2024-05-13
Update readme

chowbao:
Updates to the README


README.md
```diff
@@ -1,210 +1,119 @@
...
-### **export_config_settings**
+---
 
-```bash
-> stellar-etl export_config_settings --end-ledger 500000 --output export_config_settings.txt
-```
+## **Export Commands**
 
-Exports historical config settings data from the genesis ledger to the provided end-ledger to an output file. The command reads from the bucket list, which includes the full history of the Stellar ledger. As a result, it should be used in an initial data dump. In order to get config settings data information within a specified ledger range, see the export_ledger_entry_changes command.
+These commands export information using the Ledger Exporter output files within a specified datastore (currently [datastore](https://github.com/stellar/go/tree/master/support/datastore) only supports GCS). This allows users to provide a start and end ledger range. The commands in this category export a list of everything that occurred within the provided range. All of the ranges are inclusive.
```
sydneynotthecity, line 90: may be worthwhile to link Ledger Exporter to its README if someone wants to find out more. Or just move the note that follows below up in the readme

sydneynotthecity: Might also be worthwhile to add a note that your datastore must contain the expected XDR files to work, requiring the operator to run a Ledger Exporter to hydrate that data store.

README.md
```diff
@@ -1,210 +1,119 @@
...
+| datastore-path | Datastore bucket path to read txmeta files from                                               | ledger-exporter/ledgers |
+| buffer-size    | Buffer size sets the max limit for the number of txmeta files that can be held in memory      | 1000                    |
+| num-workers    | Number of workers to spawn that read txmeta files from the datastore                          | 5                       |
+| retry-limit    | Datastore GetLedger retry limit                                                               | 3                       |
+| retry-wait     | Time in seconds to wait for GetLedger retry                                                   | 5                       |
 
-```bash
-> stellar-etl export_ttl --end-ledger 500000 --output export_ttl.txt
-```
+> _*NOTE:*_ Using captive-core requires a Stellar Core instance that is v20.0.0 or later. The commands use the Core instance to retrieve information about changes from the ledger. These changes can be in the form of accounts, offers, trustlines, claimable balances, liquidity pools, or account signers.
```
sydneynotthecity, line 109: might be easier to link to ledger entry types in our stellar docs, versus enumerating each one individually.

README.md
```diff
@@ -1,210 +1,119 @@
...
+> As the Stellar network grows, the Stellar Core instance has to catch up on an increasingly large amount of information. This catch-up process can add some overhead to the commands in this category. In order to avoid this overhead, run prefer processing larger ranges instead of many small ones, or use unbounded mode.
```
sydneynotthecity, line 110: Is there a certain amount of RAM/CPU that must be allocated for catchup to work? We might want to include resource recommendations as well as expected runtimes to build a captive core db from scratch.

README.md
```diff
@@ -302,82 +218,80 @@ As the Stellar network grows, the Stellar Core instance has to catch up on an in
...
+#### **Unbounded**
 
-Orderbooks are exported in batches of a size defined by the `batch-size` flag. By default, the batch-size parameter is set to 64 ledgers, which corresponds to a five minute period of time. This batch size is convenient because checkpoint ledgers are created every 64 ledgers. Checkpoint ledgers act as anchoring points in that once they are available, so are the previous 63 nodes. It is beneficial to export in multiples of 64.
+If only a start ledger is provided, then the command runs in an unbounded fashion starting from the provided ledger. In this mode, stellar-etl will block and wait for the next sequentially written ledger file in the datastore. Since the changes are continually exported in batches, this process can be continually run in the background in order to avoid the overhead of closing and starting new stellar-etl instances.
```
sydneynotthecity, line 231: Do we know if unbounded mode _actually_ works? We might need to test, or add a note stating that this mode is currently unsupported.

README.md
```diff
@@ -302,82 +218,80 @@ As the Stellar network grows, the Stellar Core instance has to catch up on an in
...
-	 - If working with captive core, the methods need to work in the background. There should be methods that export batches of data and send them to a channel. There should be other methods that read from the channel and transform the data so it can be exported.
+- `export_new_data_structure.go` in the `cmd` folder
+  - This file can be generated with cobra by calling: `cobra add {command}`
+  - This file will parse flags, create output files, get the transformed data from the input package, and then export the data.
+- `export_new_data_structure_test.go` in the `cmd` folder
+  - This file will contain some tests for the newly added command. The `runCLI` function does most of the heavy lifting. All the tests need is the command arguments to test and the desired output.
+  - Test data should be stored in the `testdata/new_data_structure` folder
+- `new_data_structure.go` in the `internal/input` folder
+  - This file will contain the methods needed to extract the new data structure from wherever it is located. This may be the history archives, the bucket list, or a captive core instance.
+  - This file should extract the data and transform it, and return the transformed data.
```
sydneynotthecity, line 291: we're not actually transforming data in `internal/input`. This should move under `internal/transform`

sydneynotthecity (commented):
Couple suggestions to fix

README.md
chowbao: I think technically ledgers and assets does do a small transformation (LCM to the history archive struct for changes). But you're right we shouldn't be doing this and this should only be extracting data and the transforms should happen in the transform dir

README.md
chowbao: I'll add that it's currently unsupported

README.md
chowbao: Added our original resource allocation when we ran with captive-core

sydneynotthecity (approved):
LGTM

PR #268, opened 2024-07-16
Update the data type of AssetType in trustline tables

amishas157:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

This PR updates the data type of AssetType for Trustlines tables from integer to string. 

### Why

This is done to make AssetType consistent across all the schemas as it was mistakenly added as integer earlier.

### Known limitations

None. Posted the deploy plan in jira

### DIFF
--- a/internal/transform/trustline.go
+++ b/internal/transform/trustline.go
@@ -39,15 +39,13 @@ func TransformTrustline(ledgerChange ingest.Change, header xdr.LedgerHeaderHisto
 		return TrustlineOutput{}, errors.Wrap(err, fmt.Sprintf("could not create ledger key string for trustline with account %s and asset %s", outputAccountID, asset.ToAsset().StringCanonical()))
 	}
 
-	var assetTypeString string
 	if asset.Type == xdr.AssetTypeAssetTypePoolShare {
 		poolID = PoolIDToString(trustEntry.Asset.MustLiquidityPoolId())
-		assetTypeString = "pool_share"
+		assetType = "pool_share"
 	} else {
 		if err = asset.Extract(&assetType, &outputAssetCode, &outputAssetIssuer); err != nil {
 			return TrustlineOutput{}, errors.Wrap(err, fmt.Sprintf("could not parse asset for trustline with account %s", outputAccountID))
 		}
-		assetTypeString = assetType
 	}
 
 	outputAssetID := FarmHashAsset(outputAssetCode, outputAssetIssuer, asset.Type.String())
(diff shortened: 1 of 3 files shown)


internal/transform/trustline.go
```diff
@@ -61,7 +61,7 @@ func TransformTrustline(ledgerChange ingest.Change, header xdr.LedgerHeaderHisto
 	transformedTrustline := TrustlineOutput{
 		LedgerKey:          outputLedgerKey,
 		AccountID:          outputAccountID,
-		AssetType:          int32(asset.Type),
+		AssetType:          asset.Type.String(),
```
chowbao, line 64: nit: I think this will work but it would be nice to make the asset information consistent between the different transformations like [here](https://github.com/stellar/stellar-etl/blob/master/internal/transform/asset.go#L57-L60)
```
err := asset.Extract(&outputAssetType, &outputAssetCode, &outputAssetIssuer)
	if err != nil {
		return TrustlineOutput{}, fmt.Errorf("some error message")
	}
```

At least I think this code should have access to `Extract()`

amishas157: @chowbao Is there a reason why we are extracting assetType, code and issuer conditionally here:
https://github.com/stellar/stellar-etl/blob/0e50a2281af62fa67526bd7330557c72bce24f90/internal/transform/trustline.go#L44-L48

Can we extract this info regardless of assetType being poolshare?

chowbao: Oh oops the code already does this `Extract()`. Looking at your code it does seem like the way you wrote it is the easiest way to do this. `assetType` isn't even used in the code 😅 

The condition is needed because `xdr.AssetTypeAssetTypePoolShare` does not contain `assetCode` nor `assetIssuer`

amishas157, line 64: @chowbao I am seeing values like `AssetTypeAssetTypePoolShare` in the tables
Screenshot 2024-07-19 at 10 26 15 AM

Should it rather be `AssetTypePoolShare` ?

chowbao: It should match how `history_assets` does it which I think removes `AssetTypeAssetType`.

You might need to define a new function to make the change or look for an existing if one exists

internal/transform/trustline.go
```diff
@@ -39,12 +39,15 @@ func TransformTrustline(ledgerChange ingest.Change, header xdr.LedgerHeaderHisto
 		return TrustlineOutput{}, errors.Wrap(err, fmt.Sprintf("could not create ledger key string for trustline with account %s and asset %s", outputAccountID, asset.ToAsset().StringCanonical()))
 	}
 
+	var assetTypeString string
```
chowbao, line 42: nit: I think you don't need to create a new variable and can just reuse `assetType` because it is already a string variable defined on line 33

In any case this is a non blocking suggestion

chowbao (approved):

PR #270, opened 2024-07-24
Add json field to operations export

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

Add JSON blob of `operation.details`

### Why

To leverage BigQuery JSON data type as well as aid downstream dbt models

### Known limitations

* Need to test in test-hubble

### DIFF
--- a/internal/transform/operation.go
+++ b/internal/transform/operation.go
@@ -83,17 +83,18 @@ func TransformOperation(operation xdr.Operation, operationIndex int32, transacti
 	outputLedgerSequence := utils.GetLedgerSequence(ledgerCloseMeta)
 
 	transformedOperation := OperationOutput{
-		SourceAccount:       outputSourceAccount,
-		SourceAccountMuxed:  outputSourceAccountMuxed.String,
-		Type:                outputOperationType,
-		TypeString:          outputOperationTypeString,
-		TransactionID:       outputTransactionID,
-		OperationID:         outputOperationID,
-		OperationDetails:    outputDetails,
-		ClosedAt:            outputCloseTime,
-		OperationResultCode: outputOperationResultCode,
-		OperationTraceCode:  outputOperationTraceCode,
-		LedgerSequence:      outputLedgerSequence,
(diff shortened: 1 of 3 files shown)


internal/transform/operation.go
```diff
@@ -83,17 +83,18 @@ func TransformOperation(operation xdr.Operation, operationIndex int32, transacti
...
+		Type:                 outputOperationType,
+		TypeString:           outputOperationTypeString,
+		TransactionID:        outputTransactionID,
+		OperationID:          outputOperationID,
+		OperationDetails:     outputDetails,
+		ClosedAt:             outputCloseTime,
+		OperationResultCode:  outputOperationResultCode,
+		OperationTraceCode:   outputOperationTraceCode,
+		LedgerSequence:       outputLedgerSequence,
+		OperationDetailsJSON: outputDetails,
```
amishas157, line 97: What is the real difference between OperationDetails and OperationDetailsJSON

internal/transform/operation_test.go
sydneynotthecity, line 1: This is gonna be obnoxious to maintain. I'll start communicating the changes to external partners so we can work on dropping the extra `OperationDetailsJSON`

sydneynotthecity (approved):
LGTM

internal/transform/operation.go
amishas157: I see. Are you going to drop individual fields subsequently and keep only the json blob?

internal/transform/operation_test.go
chowbao: Yeah. Our unit tests are a pain in general unfortunately

internal/transform/operation.go
chowbao: We could drop the `details` record eventually after we talk with partners and they move off of it. Oh I guess we would also need to move off of it as well

So yes in theory the end result would be
* drop `details`
* rename `details_json` to `details`

amishas157 (approved):

PR #283, opened 2024-09-16
Hubble 526 protocol22 changes

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

Add support for Protocol 22 changes
* Largely this is supporting the new constructor params in the form of a new createContractV2 from InovkeHostFunction

* The relevant bls changes are only updating the `ContractCostType` enum with bls changes. This is automatically taken care of because stellar-etl config_settings handles `ContractCost` as map/dict-like `{type: value}`. If needed this can be made more user friendly with dbt transformations into a flattened table

* CAP60 may be moved out of P22

### Why

To support P22

### Known limitations

* Need to update/add tests
* Keep as draft until P22 core image running in testnet; Theoretically because of backwards compatibility this can be merged and run in prod but it would be nice to keep them separate until P22 changes are fully set in stone


internal/transform/operation.go
```diff
@@ -1082,26 +1059,38 @@ func extractOperationDetails(operation xdr.Operation, transaction ingest.LedgerT
...
-				details["asset"] = args.ContractIdPreimage.MustFromAsset().StringCanonical()
-			default:
-				panic(fmt.Errorf("unknown contract id type: %s", args.ContractIdPreimage.Type))
 			}
 		case xdr.HostFunctionTypeHostFunctionTypeUploadContractWasm:
 			details["type"] = "upload_wasm"
 			transactionEnvelope := getTransactionV1Envelope(transaction.Envelope)
 			details["ledger_key_hash"] = ledgerKeyHashFromTxEnvelope(transactionEnvelope)
 			details["contract_code_hash"] = contractCodeHashFromTxEnvelope(transactionEnvelope)
+		case xdr.HostFunctionTypeHostFunctionTypeCreateContractV2:
```
chowbao, line 1073: `HostFunctionTypeHostFunctionTypeCreateContractV2` is the same as `HostFunctionTypeHostFunctionTypeCreateContract` except it has `ConstructorArgs`

internal/transform/operation.go
```diff
@@ -1690,26 +1656,38 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
-				details["asset"] = args.ContractIdPreimage.MustFromAsset().StringCanonical()
-			default:
-				panic(fmt.Errorf("unknown contract id type: %s", args.ContractIdPreimage.Type))
 			}
 		case xdr.HostFunctionTypeHostFunctionTypeUploadContractWasm:
 			details["type"] = "upload_wasm"
 			transactionEnvelope := getTransactionV1Envelope(operation.transaction.Envelope)
 			details["ledger_key_hash"] = ledgerKeyHashFromTxEnvelope(transactionEnvelope)
 			details["contract_code_hash"] = contractCodeHashFromTxEnvelope(transactionEnvelope)
+		case xdr.HostFunctionTypeHostFunctionTypeCreateContractV2:
```
chowbao, line 1670: Operation cases are duplicated due to how the original stellar-etl was written

internal/transform/operation.go
```diff
@@ -1690,26 +1656,38 @@ func (operation *transactionOperationWrapper) Details() (map[string]interface{},
...
+			details["type"] = "create_contract_v2"
+
+			transactionEnvelope := getTransactionV1Envelope(operation.transaction.Envelope)
+			details["ledger_key_hash"] = ledgerKeyHashFromTxEnvelope(transactionEnvelope)
+			details["contract_id"] = contractIdFromTxEnvelope(transactionEnvelope)
+			details["contract_code_hash"] = contractCodeHashFromTxEnvelope(transactionEnvelope)
+
+			// ConstructorArgs is a list of ScVals
+			// This will initially be handled the same as InvokeContractParams until a different
+			// model is found necessary.
```
sydneynotthecity, line 1681: when you say "until a different model is found necessary" do you mean if we need to update the schema/data model? Or is there a possibility core will change how they present this data?

internal/transform/operation.go
sydneynotthecity: for `createcontractV2` will `parameters` ever be overloaded with both function parms and contructorargs? Or are we just repurposing the `parameters_decoded` for constructor args?

sydneynotthecity (commented):
To-do: fix unit tests and looks good 🤷‍♀️

internal/transform/operation.go
chowbao: > Or are we just repurposing the parameters_decoded for constructor args?

This. It uses the same []ScVal code currently. 

> for createcontractV2 will parameters ever be overloaded with both function parms and contructorargs?

One thing to clarify is that the original parameters are for the `HostFunctionTypeHostFunctionTypeInvokeContract` invokehost operation. Whereas the `HostFunctionTypeHostFunctionTypeCreateContract` never had any parameters and now `HostFunctionTypeHostFunctionTypeCreateContractV2` has constructorArgs

So the CreateContract calls will never have "params" as they are currently exclusive to the invokeContract. And therefore won't ever be overloaded in P23.

Note also that constructorArgs won't ever be in the invokeContract. The way I see CreateContract is like CreateContract is where `ScValTypeScvLedgerKeyContractInstance` contract data entries are created. CreateContractV2 allows a single constructor function as an initialization function and has [other restrictions](https://github.com/stellar/stellar-protocol/blob/master/core/cap-0058.md#design-rationale). Think of this like creating an object with 
```
class car():
  def __init__(self, make, model, color):
    self.make = make
    self.model = model
    self.color = color

  def accelerate(self):
    do_stuff
```
Any calls to the object functions no longer care about the __init__ params as it's already been initialized to the self.* of the object

internal/transform/operation_test.go
sydneynotthecity, line 1: nice

sydneynotthecity (approved):
LGTM

Is now the right time to throw a deprecation warning that we plan to drop stellar-core support from ETL directly?

chowbao:
> LGTM
> 
> Is now the right time to throw a deprecation warning that we plan to drop stellar-core support from ETL directly?

That's a good idea. I'll throw in the deprecation warning

PR #304, opened 2025-01-22
Pass through change updates to change compactor

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

[Github issue](https://github.com/stellar/go/issues/5540)

Moving a copied version of the changeCompactor with the new Change struct fields into stellar-etl to pass through transaction_id, operation_id, and operation_type to ledger entry changes

### Why

To get more useful join keys in the ledger entry change tables in BQ

### Known limitations

* transaction_id, operation_id, and operation_type may not be entirely correct because we are using compacted changes where we keep the data for the latest change within a ledger. This should be fixed when stellar-etl is moved off of using changeCompactor and the ledger entry change table saves all changes instead of compacted changes
* Some of this plus simpler logic could be used depending on https://github.com/stellar/go/issues/5578


internal/transform/account_signer.go
```diff
@@ -24,12 +25,10 @@ func TransformSigners(ledgerChange ingest.Change, header xdr.LedgerHeaderHistory
...
 	}
 
-	closedAt, err := utils.TimePointToUTCTimeStamp(header.Header.ScpValue.CloseTime)
-	if err != nil {
-		return signers, err
-	}
-
-	ledgerSequence := header.Header.LedgerSeq
+	closedAt := ledger.ClosedAt(*ledgerChange.Ledger)
+	ledgerSequence := ledger.Sequence(*ledgerChange.Ledger)
```
chowbao, line 29: XDRill magic

go.mod
```diff
@@ -1,12 +1,12 @@
 module github.com/stellar/stellar-etl
 
-go 1.22
+go 1.23
```
sydneynotthecity, line 3: yay

internal/input/change_compactor.go
```diff
@@ -0,0 +1,265 @@
+package input
```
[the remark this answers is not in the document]
sydneynotthecity: what is the reason we can't use the change_compactor from `stellar/go/ingest`? Will this be swapped out with the actual change compactor in the future?

internal/transform/schema.go
```diff
@@ -126,6 +126,8 @@ type AccountSignerOutput struct {
 	Deleted            bool        `json:"deleted"`
 	ClosedAt           time.Time   `json:"closed_at"`
 	LedgerSequence     uint32      `json:"ledger_sequence"`
+	TransactionID      int64       `json:"transaction_id"`
+	OperationID        int64       `json:"id"`
```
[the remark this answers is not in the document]
sydneynotthecity: potentially `OperationType`? I think there could be merit if we think we're going to join frequently to `history_operations` just to get the operation type. Otherwise, I can't think of anything else.

Also nit: is `id` consistent with our other tables and what we want `OperationID` to be called? it's a pretty generic field name

internal/input/change_compactor.go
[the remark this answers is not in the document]
chowbao: Ah this might not have been captured outside of the horizon/cdp planning meeting. We agreed not to add propagate the changes through stellar/go/ingest/change_compactor because it gives imperfect data that is specific for stellar-etl's use case. Because of that we need to fork the change_compactor changes to use the new Change fields in stellar-etl

Options were to fork/branch all of stellar/go, make the changeCompactor changes and point to that fork/branch
or
copy and paste the change_compactor directly into stellar-etl with the changes we needed instead.

I opted for the second option with the theory that it'll be easier to maintain as we can continue to pull in new stellar/go change independently from our own fork/branch of change_compactor

internal/transform/schema.go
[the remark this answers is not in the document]
chowbao: Yeah let's add in `OperationType` for now. We can drop it if we don't find it useful at a later date

> Also nit: is id consistent with our other tables and what we want OperationID to be called? it's a pretty generic field name

The OperationID is consistent. It uses the same toid(ledger sequence, transaction index, option index + 1) logic. Oh you're right the `json:id` should be renamed. I copy and pasted the schema from operationOutput{} and didn't clarify the `json:id` field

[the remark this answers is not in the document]
chowbao: Added `OperationType`
Renamed `json:id` to `json:operation_id`

updated in ac1ca80bd30312dd7a00ee6109ab02b7a35e04df

internal/utils/main.go
```diff
@@ -1101,3 +1104,45 @@ type HistoryArchiveLedgerAndLCM struct {
...
+
+type ChangeDetails struct {
+	ClosedAt       time.Time
+	LedgerSequence uint32
+	OperationType  null.Int
+	TransactionID  null.Int
+	OperationID    null.Int
+}
+
+func GetChangesDetails(ledgerChange ingest.Change) ChangeDetails {
```
chowbao, line 1116: @sydneynotthecity this is what it'll take to get the ChangeDetails from `ingest.Change`

Honestly not the worst thing we've had to do to get data but definitely could be optimized more with a few tweaks to `ingest.Change` instead of doing all this logic in the application

One important note (also left as a comment in the code), the TransactionID, OperationID, and OperationType may be incorrect as the change has been compacted. It is possible for a compacted change to have had pointed to more than 1 operation and this logic will just choose the latest available TransactionIndex and OperationIndex

chowbao:
One other important note. This requires a schema change for the raw tables. I think it'd be worthwhile to have this run in the test airflow over the weekend if possible

sydneynotthecity:
> I think it'd be worthwhile to have this run in the test airflow over the weekend if possible

Yup, agreed. I also think we should modify the dune exports and drop these new columns and determine whether or not it's necessary to propagate into dune. If our end goal is to create a `token_transfers` table there, i don't think it's worthwhile to add to external facing warehouses.

internal/input/change_compactor.go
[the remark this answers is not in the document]
sydneynotthecity: We discussed offline, but let's subscribe to new issues/bugs that are opened in the go monorepo. I agree this is better than us working off a fork. But this does introduce risk in our code losing parity from `main`, which is especially risky if there are ever bugs squashed in the original change compactor. I think this is low risk but something to still watch.

internal/transform/config_setting.go
```diff
@@ -143,8 +138,11 @@ func TransformConfigSetting(ledgerChange ingest.Change, header xdr.LedgerHeaderH
 		LastModifiedLedger:              uint32(ledgerEntry.LastModifiedLedgerSeq),
 		LedgerEntryChange:               uint32(changeType),
 		Deleted:                         outputDeleted,
-		ClosedAt:                        closedAt,
-		LedgerSequence:                  uint32(ledgerSequence),
+		ClosedAt:                        changeDetails.ClosedAt,
+		LedgerSequence:                  changeDetails.LedgerSequence,
+		TransactionID:                   changeDetails.TransactionID,
+		OperationID:                     changeDetails.OperationID,
+		OperationType:                   changeDetails.OperationType,
```
sydneynotthecity, line 145: i think we can actually drop transaction id, operation id and type from this table. config settings are only an internal change, configured by validators. my assumption is that these three fields will be `nil`

internal/transform/trade_test.go
```diff
@@ -809,7 +809,7 @@ func makeTradeTestOutput() [][]TradeOutput {
 		LiquidityPoolFee:       null.IntFrom(30),
 		HistoryOperationID:     101,
 		TradeType:              2,
-		RoundingSlippage:       null.IntFrom(100),
+		RoundingSlippage:       null.IntFrom(9),
```
sydneynotthecity, line 812: just curious - was there a reason to update this test?

internal/utils/main.go
```diff
@@ -1101,3 +1104,45 @@ type HistoryArchiveLedgerAndLCM struct {
...
+	var outputOperationType null.Int
+	var outputTransactionID null.Int
+	var outputOperationID null.Int
+	var lcm *xdr.LedgerCloseMeta
+
+	if ledgerChange.Ledger == nil {
+		lcm = &ledgerChange.Transaction.Ledger
+	} else {
+		lcm = ledgerChange.Ledger
+	}
```
sydneynotthecity, line 1126: little bit of a weird redundancy that you can access the lcm twice within `ingest.Change`. Not pertinent to this Pr, but i wonder if it makes more sense just to always pass the lcm since you could access the txn from there?

internal/utils/main.go
```diff
@@ -1101,3 +1104,45 @@ type HistoryArchiveLedgerAndLCM struct {
...
+	closedAt := ledger.ClosedAt(*lcm)
+	ledgerSequence := ledger.Sequence(*lcm)
+	if ledgerChange.Transaction != nil {
+		outputTransactionID = null.NewInt(toid.New(int32(ledgerSequence), int32(ledgerChange.Transaction.Index), 0).ToInt64(), true)
+		// outputOperationID may point to the incorrect operation because the change compactor takes the latest OperationIndex if it exists
+		// It is possible that OperationIndex is 0 and invalid but will be used to populate outputOperationID anyways
+		// TODO: This should be fixed when the LedgerEntryChange transforms/processors use the actual changes instead of compacted changes
+		outputOperationID = null.NewInt(toid.New(int32(ledgerSequence), int32(ledgerChange.Transaction.Index), int32(ledgerChange.OperationIndex+1)).ToInt64(), true)
+		operation, ok := ledgerChange.Transaction.GetOperation(ledgerChange.OperationIndex)
```
sydneynotthecity, line 1135: Just confirming: there is no way that the OperationType will get out of sync with the OperationIndex? It's not possible to cross them and have a mismatched OperationID <> OperationType, correct?

sydneynotthecity (commented):
Looks good, i didn't review the golden files and just assume they implicitly work based on tests passing

internal/utils/main.go
chowbao: Yup I agree. [Issue](https://github.com/stellar/go/issues/5578) for always passing LCM

internal/transform/trade_test.go
chowbao: Yes there was an [orderbook bug fix](https://github.com/stellar/go/pull/5541) that was pulled in with the latest stellar/go changes. I think our original test was passing under that bug

internal/utils/main.go
chowbao: Issue is now approved and pointed

Issue #298, opened 2025-01-02
Asset Type is incorrect enum in `contract_data`


sydneynotthecity:
While working on data quality tests, I found that `assetType` has following values in `stg_contract_data`

- AssetTypeAssetTypeCreditAlphanum4

Acceptable values are 'credit_alphanum4', 'credit_alphanum12', 'native' 

 

Ref: https://stellarfoundation.slack.com/archives/C071DNLU4JC/p1728081984293059

This ticket requires to fix the code creating these values and backfill the existing data. Note: if we do decide to fix this bug in the raw transformation layer, we will need to coordinate backfills with Nansen and Dune, who use Hubble as their source of record. It may not be ideal to fix in the raw, and might be worth mapping correctly within dbt instead.

PR #331, opened 2025-06-09
Add base64 key hash and correct asset type to contracts

chowbao:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

https://github.com/stellar/stellar-etl/issues/298
https://github.com/stellar/stellar-dbt/issues/301

Fixing asset_type in contract_data
Adding base64 ledgerKeyHash to contract_code and contract_data (note that the ttl key_hash type xdr.Hash cannot be converted back into xdr.LedgerKey type)

### Why

Fixing bugs in stellar-etl and making ledger key entry matching work with base64 keys

### Known limitations

* ttl.KeyHash is saved as xdr.Hash and xdr.MarshallBase64() on it will result in different a different output compared to like rpc.getLedgerEntries. We should keep the original keyHash and ledgerKeyHash within Hubble so we can associate contract_data/code with ttl. A new LedgerKeyHashBase64 column is added to contract_data/code that is the same as the getLedgerEntries ledger key xdr
* Use UDF to backfill or just fix with the Feb 2024 to current smart contract backfill run


chowbao:
> > note that the ttl key_hash type xdr.Hash cannot be converted back into xdr.LedgerKey type
> 
> Is it problematic from backfill perspective? Can you give some more context?
> Did not understand constraints about ttl. Otherwise, looks good to me.

It is not. Backfilling will only change the new base64 key_hash column.

Currently key_hashes are saved like `66d50c209f3a3d7fa3b1dbdf7bd165dfc593bc73027e263629e6ee5a7649c442` (hex string)

The base64 will look more like `AAAAEwAAAAAn13idybgZn5YtZNWqyccASQFly4n8+hf`

The problem with ttl.KeyHash is that it is of type xdr.Hash instead of xdr.LedgerKey and has already gone through "hashing/sha" like [here](https://github.com/stellar/stellar-rpc/blob/6bb9d7a387d51e25c0e0d2c3417e31da78312f68/cmd/stellar-rpc/internal/preflight/preflight_test.go#L228)

This means that the marshall/conversion of the xdr.LedgerKey of say a contact_data entry would have a different result compared to marshall/conversion of the xdr.Hash. So we need to apply the hash/sha to xdr.LedgerKey to make it equal to the ttl.KeyHash

Issue #381, opened 2026-02-26
Fix resource_fee_refund logic in stellar-etl


chowbao:
[Slack thread](https://stellarfoundation.slack.com/archives/C073MNCNVQQ/p1772068194582229?thread_ts=1772044535.691019&cid=C073MNCNVQQ)

As seen from the [Stellar XDR changes](https://github.com/stellar/stellar-xdr/blob/curr/Stellar-ledger.x#L558-L570), the fee ledgerEntryChange has moved to a new section within the transaction result struct since Protocol 23.

We are currently not accounting for this new `postTxApplyFeeProcessing` part of the struct when calculating the account start/end balances within a transaction which is used to calculate the `resource_fee_refund`

Because this change is only for P23 the change only applies to transaction meta v4 we only need to make changes to that section of the code.

In theory, all we have to add is an append of the `metav4.TxChangesAfter` and `transaction.PostTxApplyFeeChanges`. We should test and compare with stellar.expert to make sure the resource fee refund is correct. We can also add an invariant check to the code itself or downstream dbt tests

```golang
  accountBalanceStart, accountBalanceEnd := getAccountBalanceFromLedgerEntryChanges(append(metav4.TxChangesAfter, transaction.PostTxApplyFeeChanges), feeAccountAddress)
  outputResourceFeeRefund = accountBalanceEnd - accountBalanceStart
```

TBD
We can include the backfill of the history_transaction table from P23 to now in this ticket or the backfill can be a separate ticket. The preference for backfilling would be to work with whomever is on-call that week to do the backfill

* Backfill history_transactions
* Update stg_history_transactions
* See if anything downstream of stg_history_transactions actually uses the refund fee. If so then backfill that too
* Determine if we want to do a backfill for Dune as well
* ???
* Profit

Alternative

Just an FYI there are functions we wrote (xdrill helper functions) that does have this correct change [here](https://github.com/stellar/go-stellar-sdk/blob/main/ingest/ledger_transaction.go#L624C29-L649). My preference right now is to just make the hard coded change above instead of refactoring the transform logic to use the xdrill helper functions.

ttboogie:
Need to inform dune of this backfill - when we rollup all updates to give them

PR #383, opened 2026-03-03
resource_fee_refund update for P23+, and test with soroban fee bump txn

ttboogie:
### What

Updating `transform/transaction.go` to ingest `resource_fee_refund` accurately for P23+ txn.

### Why

Previously, `resource_fee_refund` has been 0.

### Known limitations

N/A


amishas157 (commented):
The changes look good to me. Can you also verify the value of resource_refund_fee for some of the transactions in stellar expert like suggested in ticket https://github.com/stellar/stellar-etl/issues/381#issue-3997448603 

> We should test and compare with stellar.expert to make sure the resource fee refund is correct.

PR #401, opened 2026-04-06
initial p26 updates

hunterpack:
<details>

### PR Structure

      otherwise).

### Thoroughness

### Release planning

</details>

### What

[TODO: Short statement about what is changing.]

### Why

[TODO: Why this change is being made. Include any context required to understand the why.]

### Known limitations

[TODO or N/A]


internal/transform/config_setting.go
```diff
@@ -191,3 +235,69 @@ func serializeParams(costParams xdr.ContractCostParams) []map[string]string {
...
+// serializeEncodedLedgerKeys converts a slice of EncodedLedgerKey (opaque bytes) to a JSON array of base64 strings.
+func serializeEncodedLedgerKeys(keys []xdr.EncodedLedgerKey) string {
+	if len(keys) == 0 {
+		return ""
+	}
+	encoded := make([]string, 0, len(keys))
+	for _, key := range keys {
+		encoded = append(encoded, base64.StdEncoding.EncodeToString(key))
+	}
+	result, _ := json.Marshal(encoded)
```
chowbao: Instead of a serialized string we might be better off as an array of strings (`[]string`).
And instead of encoding and marshaling like above you can use the [xdr.MarshalBase64](https://github.com/stellar/go-stellar-sdk/blob/main/xdr/main.go#L147-L149)

An example of this can be found [here](https://github.com/stellar/stellar-etl/blob/master/internal/transform/ledger.go#L213-L225)

We can check if we are doing the marshaling and conversions correctly because these should all be real ledger keys as in we should be able to see the ledger key for the contract data or trustline or whatever like
```
Contract Data Ledger keys
* ABC123
* EFG456

Frozen Ledger Keys
* ABC123 -- this ledger key should be the same format as the ledger key above
```

internal/transform/config_setting.go
```diff
@@ -106,6 +109,34 @@ func TransformConfigSetting(ledgerChange ingest.Change, header xdr.LedgerHeaderH
...
+	nominationTimeoutIncrementMilliseconds := contractScpTiming.NominationTimeoutIncrementMilliseconds
+	ballotTimeoutInitialMilliseconds := contractScpTiming.BallotTimeoutInitialMilliseconds
+	ballotTimeoutIncrementMilliseconds := contractScpTiming.BallotTimeoutIncrementMilliseconds
+
+	// P26 CAP-77: Frozen ledger keys (IDs 17-20)
+	frozenLedgerKeys, _ := configSetting.GetFrozenLedgerKeys()
+	frozenLedgerKeysJSON := serializeEncodedLedgerKeys(frozenLedgerKeys.Keys)
+
+	frozenLedgerKeysDelta, _ := configSetting.GetFrozenLedgerKeysDelta()
+	frozenLedgerKeysDeltaJSON := serializeFrozenKeysDelta(frozenLedgerKeysDelta)
```
chowbao: I'd recommend flattening this out so that you just have like `frozenLedgerKeysToFreeze` and `frozenLedgerKeysToUnfreeze` instead of a new map to hold both entries

For example in BigQuery this would just look like

```sql
select
  frozen_ledger_keys_to_freeze
  , frozen_ledger_keys_to_unfreeze
```

vs

```sql
select
  frozen_ledger_keys_delta.keys_to_freeze
  , frozen_ledger_keys_delta.keys_to_unfreeze
```

internal/transform/config_setting.go
```diff
@@ -106,6 +109,34 @@ func TransformConfigSetting(ledgerChange ingest.Change, header xdr.LedgerHeaderH
...
+
+	freezeBypassTxs, _ := configSetting.GetFreezeBypassTxs()
+	freezeBypassTxsJSON := serializeHashes(freezeBypassTxs.TxHashes)
+
+	freezeBypassTxsDelta, _ := configSetting.GetFreezeBypassTxsDelta()
+	freezeBypassTxsDeltaJSON := serializeFreezeBypassTxsDelta(freezeBypassTxsDelta)
```
chowbao: Same recommendation as https://github.com/stellar/stellar-etl/pull/401/changes#r3041684008

Flatten instead of making it a map

internal/transform/config_setting.go
```diff
@@ -191,3 +235,69 @@ func serializeParams(costParams xdr.ContractCostParams) []map[string]string {
...
+	if len(hashes) == 0 {
+		return ""
+	}
+	encoded := make([]string, 0, len(hashes))
+	for _, h := range hashes {
+		encoded = append(encoded, hex.EncodeToString(h[:]))
+	}
+	result, _ := json.Marshal(encoded)
+	return string(result)
+}
```
chowbao: Similar to https://github.com/stellar/stellar-etl/pull/401/changes#r3041658556

I _think_ you can convert the hash to a transaction_hash like [here](https://github.com/stellar/stellar-etl/blob/master/internal/transform/transaction.go#L22)

We would need to confirm the frozen/unfrozen tx hashes match actual tx hashing or whatever hashes == hashes we are comparing

internal/transform/schema.go
```diff
@@ -619,6 +619,19 @@ type ConfigSettingOutput struct {
...
+	LedgerTargetCloseTimeMilliseconds      uint32              `json:"ledger_target_close_time_milliseconds"`
+	NominationTimeoutInitialMilliseconds   uint32              `json:"nomination_timeout_initial_milliseconds"`
+	NominationTimeoutIncrementMilliseconds uint32              `json:"nomination_timeout_increment_milliseconds"`
+	BallotTimeoutInitialMilliseconds       uint32              `json:"ballot_timeout_initial_milliseconds"`
+	BallotTimeoutIncrementMilliseconds     uint32              `json:"ballot_timeout_increment_milliseconds"`
+	// P26 CAP-77 frozen ledger keys
+	FrozenLedgerKeys                       string              `json:"frozen_ledger_keys"`
+	FrozenLedgerKeysDelta                  string              `json:"frozen_ledger_keys_delta"`
+	FreezeBypassTxs                        string              `json:"freeze_bypass_txs"`
+	FreezeBypassTxsDelta                   string              `json:"freeze_bypass_txs_delta"`
```
chowbao: Update these to `[]string` to work the the above suggested changes

chowbao (approved):

PR #412, opened 2026-04-18
Use GCS datastore backend for get_ledger_range_from_times

chowbao:
## Summary
- Swaps the `get_ledger_range_from_times` binary search from the Stellar history archive to the GCS datastore used by the bulk export commands, so the CLI no longer depends on `historyarchive.GetLedgerHeader` for this lookup.
- Preserves the old adaptive-step implementation as `GetLedgerRangeHistoryArchive` in `internal/input/ledger_range_history_archive.go` for reference / programmatic use.
- Adds a `--datastore-path` flag (defaults to `sdf-ledger-close-meta/v1/ledgers`) so callers can point at a different datastore layout.
- Adds unit tests for `pointAt` (fetch, cache, GetFile error, decode error) and the `findLedgerForTime` boundary predicate, including a datastore-error propagation case.

## Test plan
- [x] `go build ./...`
- [x] `go test -v -run 'LedgerFinder' ./internal/input/...`
- [x] `pre-commit run --files ...` (golangci-lint clean)
- [x] Run `make int-test` locally against a datastore bucket to confirm parity with the history-archive implementation on a known time range
- [x] Sanity-check CLI: `./stellar-etl get_ledger_range_from_times --start-time ... --end-time ... --datastore-path sdf-ledger-close-meta/v1/ledgers`

### DIFF
--- a/cmd/get_ledger_range_from_times.go
+++ b/cmd/get_ledger_range_from_times.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/stellar/stellar-etl/v2/internal/input"
+	"github.com/stellar/stellar-etl/v2/internal/utils"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


hunterpack (approved):
lgtm

PR #413, opened 2026-04-20
Batch-stream history export commands

chowbao:
## Summary

Convert the nine history export commands from "load full range → write single file → upload" to per-batch streaming that matches the existing `export_ledger_entry_changes` pattern:

- Add `input.StreamLedgerBatches` and `LedgerBatch`: a goroutine-driven stream of pre-fetched LCMs in batch-size increments. Uses inclusive boundaries so single-ledger ranges emit one batch (`StreamChanges` skips them).
- Add per-ledger helpers (`TransactionsFromLedger`, `OperationsFromLedger`, `TradesFromLedger`, `PaymentOperationsFromLedger`, `HistoryArchiveLedgerFromLCM`) alongside the existing `Get*` functions.
- Add `utils.AddHistoryArchiveFlags` and `MustHistoryArchiveFlags`: folder output with `--batch-size`, replacing `--output` (single file) + `--limit`.
- Refactor `export_transactions`, `export_operations`, `export_effects`, `export_trades`, `export_assets`, `export_contract_events`, `export_token_transfers`, `export_ledgers`, `export_ledger_transaction` to stream batches, write `{start}-{end}-<type>.txt` per batch, and upload each batch before requesting the next. `export_assets` dedup now persists across batches.
- Unify captive-core and datastore paths in `export_ledgers` and `export_assets` via `utils.CreateLedgerBackend`; `--captive-core` now starts real captive core instead of using the history-archive backend (captive-core is already deprecated post-Protocol 23).
- Tests: folder `-o`, `SortForComparison` on all cases, drop `--limit` "range too large" tests, add one `-b` multi-batch test per command. Golden files regenerated via `make int-test-update`.

## Test plan

- [ ] `go test -v -cover ./internal/transform` passes
- [ ] `golangci-lint run` clean
- [ ] `make int-test` passes against regenerated golden files
- [ ] Spot-check a multi-batch run produces `{start}-{end}-<type>.txt` files and uploads each batch incrementally
- [ ] Confirm single-ledger range still emits one batch (inclusive-boundary behavior)
- [ ] Verify `--captive-core` path on `export_ledgers` / `export_assets` starts real captive core


chowbao:
Integration test golden files were updated because of the new output grouping because all export_* commands will now dump to a directory with whatever set batch sizes (default is 64 ledgers per file).

In theory we can lump all the commands into a single export file similar to how export_ledger_entry_changes does right now but I think it makes more sense to keep most export_* commands separate even if they all take the same LCM input.

cmd/export_contract_events.go
```diff
@@ -13,67 +14,50 @@ import (
...
-		// TODO: https://stellarorg.atlassian.net/browse/HUBBLE-386 GetEnvironmentDetails should be refactored
-		commonArgs := utils.MustCommonFlags(cmd.Flags(), cmdLogger)
-		env := utils.GetEnvironmentDetails(commonArgs)
+		runLedgerBatchExport(cmd, "contract_events", new(transform.ContractEventOutputParquet), processContractEvents)
+	},
+}
 
-		transactions, err := input.GetTransactions(cmdArgs.StartNum, cmdArgs.EndNum, cmdArgs.Limit, env, cmdArgs.UseCaptiveCore)
+func processContractEvents(lcm xdr.LedgerCloseMeta, env utils.EnvironmentDetails, outFile *os.File, writeParquet bool, extra map[string]string) ([]transform.SchemaParquet, int, int) {
+	txInputs, err := input.TransactionsFromLedger(lcm, env.NetworkPassphrase)
```
amishas157, line 26: If there are failures while reading the data from ledger, it will silent pass and no data will be written. I believe we should shutdown the application if input fails.

It applies for other export commands as well

cmd/export_contract_events.go
```diff
@@ -13,67 +14,50 @@ import (
...
+	var rows []transform.SchemaParquet
+	attempts, failures := 0, 0
+	for _, txInput := range txInputs {
+		attempts++
+		events, err := transform.TransformContractEvent(txInput.Transaction, txInput.LedgerHistory)
 		if err != nil {
-			cmdLogger.Fatal("could not read transactions: ", err)
+			ledgerSeq := txInput.LedgerHistory.Header.LedgerSeq
+			cmdLogger.LogError(fmt.Errorf("could not transform contract events for transaction %d in ledger %d: %v", txInput.Transaction.Index, ledgerSeq, err))
+			failures++
```
amishas157, line 39: I see that we were continuing data processing in case of failures during transforms.  It was also the case before this PR. I wonder why we do not fail in case of failed transforms? Do we expect certain records transform to fail?

cmd/ledger_batch_export.go
```diff
@@ -0,0 +1,96 @@
...
+	if err := backend.PrepareRange(ctx, ledgerbackend.BoundedRange(startNum, commonArgs.EndNum)); err != nil {
+		cmdLogger.Fatal("could not prepare ledger range: ", err)
+	}
+
+	batchChan := make(chan input.LedgerBatch)
+	go input.StreamLedgerBatches(&backend, startNum, commonArgs.EndNum, batchSize, batchChan, cmdLogger)
+
+	totalAttempts, totalFailures := 0, 0
+	for batch := range batchChan {
+		path := filepath.Join(outputFolder, exportFilename(batch.BatchStart, batch.BatchEnd+1, exportName))
```
amishas157, line 74: The batch is inclusive until BatchEnd. I think it could be confusing to have batch.BatchEnd+1 in the file name. It should be batch.BatchEnd

amishas157:
There is a lot of dead code after this change. Might be worth cleaning up to reduce any vulnerability surface area.

  internal/utils/main.go:
  - AddArchiveFlags — replaced by AddLedgerBatchFlags. No callers.
  - MustArchiveFlags — replaced by MustLedgerBatchFlags. No callers.
  - CreateBackend — only called by the dead Get*HistoryArchive input fns below.
  - historyArchiveBackend type + all its methods (GetLatestLedgerSequence, GetLedgers, GetLedgerArchive, GetLedger, PrepareRange, IsPrepared, Close) — only constructed by
  CreateBackend.

  internal/input/:
  - GetPaymentOperations (assets.go:43) — replaced by PaymentOperationsFromLedger. No callers.
  - GetLedgers (ledgers.go:69) — no callers.
  - GetOperations (operations.go:63) — replaced by OperationsFromLedger. No callers.
  - GetTrades (trades.go:63) — replaced by TradesFromLedger. No callers.
  - GetTransactions (transactions.go:51) — replaced by TransactionsFromLedger. No callers.
  - GetPaymentOperationsHistoryArchive (assets_history_archive.go:13) — entire file dead.
  - GetLedgersHistoryArchive (ledgers_history_archive.go:10) — entire file dead.
  - GetAllHistory + AllHistoryTransformInput (all_history.go) — no callers anywhere. Pre-existing dead code that this PR doesn't touch, but worth clearing at the same time.

golden files:
  testdata/assets/large_range_assets.golden
  testdata/effects/large_range_effects.golden
  testdata/ledgers/large_range_ledgers.golden
  testdata/operations/large_range_ops.golden
  testdata/orderbooks/large_range_orderbooks.golden
  testdata/trades/large_range_trades.golden
  testdata/transactions/large_range_txs.golden

cmd/export_contract_events.go
chowbao: Yeah this is what we did before. I don't remember the original reason for not failing out. But yeah we could just make the update here to not continue one failures

cmd/export_contract_events.go
chowbao: Same as https://github.com/stellar/stellar-etl/pull/413#discussion_r3120026890

Originally it did but with this change we can make failures explicitly stop

cmd/ledger_batch_export.go
chowbao: Hmm this is how the current export_ledger_entry_changes outputs I think. Let me double check. I'll make this match with whatever export_ledger_entry_changes output does right now

chowbao:
> There is a lot of dead code after this change. Might be worth cleaning up to reduce any vulnerability surface area.
> 
> internal/utils/main.go:
> 
> * AddArchiveFlags — replaced by AddLedgerBatchFlags. No callers.
> * MustArchiveFlags — replaced by MustLedgerBatchFlags. No callers.
> * CreateBackend — only called by the dead Get*HistoryArchive input fns below.
> * historyArchiveBackend type + all its methods (GetLatestLedgerSequence, GetLedgers, GetLedgerArchive, GetLedger, PrepareRange, IsPrepared, Close) — only constructed by
>   CreateBackend.
> 
> internal/input/:
> 
> * GetPaymentOperations (assets.go:43) — replaced by PaymentOperationsFromLedger. No callers.
> * GetLedgers (ledgers.go:69) — no callers.
> * GetOperations (operations.go:63) — replaced by OperationsFromLedger. No callers.
> * GetTrades (trades.go:63) — replaced by TradesFromLedger. No callers.
> * GetTransactions (transactions.go:51) — replaced by TransactionsFromLedger. No callers.
> * GetPaymentOperationsHistoryArchive (assets_history_archive.go:13) — entire file dead.
> * GetLedgersHistoryArchive (ledgers_history_archive.go:10) — entire file dead.
> * GetAllHistory + AllHistoryTransformInput (all_history.go) — no callers anywhere. Pre-existing dead code that this PR doesn't touch, but worth clearing at the same time.
> 
> golden files: testdata/assets/large_range_assets.golden testdata/effects/large_range_effects.golden testdata/ledgers/large_range_ledgers.golden testdata/operations/large_range_ops.golden testdata/orderbooks/large_range_orderbooks.golden testdata/trades/large_range_trades.golden testdata/transactions/large_range_txs.golden

These were left on purpose. Instead of straight overwriting the original code I thought it'd be useful to keep the old functions around and then delete them in the future after the new changes are running for a week or so

cmd/export_contract_events.go
chowbao: Ah it looks like we wanted local processing to keep processing for debugging purposes instead of blocking on a single failure.

The reason why we error out in airflow is cause we set `--strict-export` to true. We we can actually make the default `--strict-export` and if running locally we want to continue on failures we can set strict-export to false

amishas157: Ah okay, that makes sense. I am okay to leave it as is.

cmd/export_contract_events.go
amishas157: Originally we were failing actually for input erroring https://github.com/stellar/stellar-etl/blob/427d2e2565c8cc98c7a2fbc65305a314c114aa33/cmd/export_ledgers.go#L33-L35

cmd/ledger_batch_export.go
chowbao: export_ledger_entry_changes has always done it this way so I think for now we can keep all the output/start-end-* the same

cmd/export_contract_events.go
amishas157: Actually you wont need to make change. For strict export, logError is equivalent to fatal
https://github.com/stellar/stellar-etl/blob/427d2e2565c8cc98c7a2fbc65305a314c114aa33/internal/utils/logger.go#L17-L23

cmd/ledger_batch_export.go
amishas157: ah okay

amishas157 (approved):

cmd/export_contract_events.go
chowbao: Updated in f118161142a7ffd52e8db2993690f068b145e643

`--strict-export true` is now the default

Also had claude add test cases for the fast fail paths

PR #423, opened 2026-05-06
Harden Release Drafter workflow

amishas157:
## Summary

Migrate the deprecated `::set-output` workflow command to `$GITHUB_OUTPUT`, and pass branch / version values through `env:` rather than re-interpolating them into bash. This:

- Removes the script-injection surface from PR-controlled values flowing into `run:` blocks (matches GitHub's [Actions hardening guide][1]).
- Unblocks future runner-image upgrades that drop `::set-output` support.

Behavior is unchanged — the workflow still produces the same `BRANCH`, `TAG`, and `NEXT_VERSION` step outputs and uses them in the same downstream steps.

Impact today is limited (workflow only runs on merged PRs from upstream branches), but the fix is mechanical defense-in-depth.

## Test plan

- [ ] After merge, the next merged PR triggers Release Drafter and tags + publishes a release as before.

[1]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,37 +31,42 @@ jobs:
 
       - name: Get branch name
         id: getbranch
-        run: echo ::set-output name=BRANCH::${GITHUB_HEAD_REF}
+        env:
+          HEAD_REF: ${{ github.head_ref }}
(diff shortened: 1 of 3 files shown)


hunterpack (approved):
lgtm

PR #446, opened 2026-07-03
Restructure agent docs: critical rules first + skill routing

chowbao:
## What
Restructures `.claude/CLAUDE.md` so the agent reads the critical safety rules first and can route to the right skill.

## Changes
- Lead with a **Critical rules** section (never run the `prod` target locally; never write to `crypto-stellar`/`hubble-261722`; run `dbt deps` after a pin/branch change; use the `bq` CLI for int64 IDs).
- Add a **task → skill** routing table pointing at the shared `.claude/skills/`.
- Fold the trailing `Rules` section into the above; minor accuracy fixes.

Docs-only — no runtime or release impact.


hunterpack (approved):
lgtm, fix pre-commit

